PR #6, opened 2019-01-04
Remove dart:mirrors usage

katis:
Mirrors are not supported in Flutter or recommended on the web platform, as reflection makes tree shaking the code impossible. (see https://www.dartlang.org/articles/server/reflection-with-mirrors).

I just removed the "overload" of reaction, as the parameter can easily be ignored by using `_` as the name.


pavanpodila:
Thanks @katis. This is great. Was thinking of doing this in the morning. Looks like you read my mind :-)

PR #8, opened 2019-01-04
More analysis rules

katis:
I added almost all [Dart linter rules](http://dart-lang.github.io/linter/lints/) and fixed all the errors (thank god for quick fix). The linter did find some actual issues with the code, which is always nice.

Had to do a bit of changes in the Interceptable/Listenable interfaces, as they needed a generic argument to type well. Also removed the "isComputed" function, as it does seem to just fine with `if (value is Computed)`, and you get the automatic casting inside the if.

I know this is a huge change, but I figured that since the test coverage seems good, and most changes are mechanical, this would be fine as one-time deal :sweat_smile:


lib/src/core/atom_derivation.dart
```diff
@@ -22,76 +16,81 @@ class Atom {
...
+      _addListener(_ListenerKind.onBecomeObserved, fn);
+
+  void Function() onBecomeUnobserved(Function fn) =>
+      _addListener(_ListenerKind.onBecomeUnobserved, fn);
 
-  void Function() _addListener(String key, Function fn) {
+  void Function() _addListener(_ListenerKind kind, Function fn) {
     if (fn == null) {
-      throw MobXException('${key} handler cannot be null');
+      throw MobXException('$kind handler cannot be null');
```
pavanpodila: I think using `$kind` might just show the number `0` or `1` and may not be that helpful. This was the reason for choosing explicit strings. Is there a quick enum -> string conversion possible for a better exception message?

lib/src/core/atom_derivation.dart
```diff
@@ -100,24 +99,24 @@ class Atom {
 enum DerivationState {
   // before being run or (outside batch and not being observed)
   // at this point derivation is not holding any data about dependency tree
-  NOT_TRACKING,
+  notTracking,
```
pavanpodila: Is that the convention for enum values in dart? I only picked up dart few weeks back but a long time JavaScript/TypeScript/C#/Java/Swift programmer ;-)

lib/src/core/atom_derivation.dart
katis: This would be formatted as  "_ListenerKind.onBecomeObserved handler cannot be null". Not ideal, but enum toString is not just the number in Dart.

pavanpodila: That is good enough for now. Not sure if we can override `toString()` for an enum

lib/src/core/atom_derivation.dart
katis: Yeah, never seen it in other languages, but for example Flutter code base uses this convention

pavanpodila:
That was a MASSIVE change @katis 🥇

lib/src/core/atom_derivation.dart
katis: Unfortunately no, enums can't have methods or anything. They're quite crippled compared to Java enum classes etc.

pavanpodila: Ya that would been a great addition to the language. Hoping it would come up in future versions of dart

PR #12, opened 2019-01-05
Inject ReactiveContext as constructor argument

katis:
Made the ReactiveContext an explicit argument for classes that need it.

Builder functions provide the global context if it's not provided for them as an optional named argument.

I tried to implement the Zone local value for the context, but it did not work properly with tests. I don't know why exactly. It can be implemented in the future by changing `mobxContext` global to a getter that finds the current Zones context.

### DIFF
--- a/lib/src/api/context.dart
+++ b/lib/src/api/context.dart
@@ -1,0 +1,4 @@
+import 'package:mobx/src/core/context.dart';
+
+/// Global Mobx context.
+final ReactiveContext mobxContext = ReactiveContext();
--- a/lib/src/api/context.dart
+++ b/lib/src/api/context.dart
@@ -1,4 +1,6 @@
 import 'package:mobx/src/core/context.dart';
 
-/// Global Mobx context.
-final ReactiveContext mobxContext = ReactiveContext();
+/// Current Mobx context.
+/// At the moment it is a singleton, but in the future
+/// it might be replaced with a Zone specific value.
+final ReactiveContext currentContext = ReactiveContext();
--- a/lib/src/api/observable.dart
+++ b/lib/src/api/observable.dart
@@ -1,3 +1,4 @@
+import 'package:mobx/src/api/context.dart';
 import 'package:mobx/src/core/atom.dart';
 import 'package:mobx/src/core/computed.dart';
 import 'package:mobx/src/core/context.dart';
@@ -17,8 +18,9 @@ import 'package:mobx/src/core/observable.dart';
 ///
 /// print('x = ${x.value}'); // read an observable's value
 /// ```
-ObservableValue<T> observable<T>(T initialValue, {String name}) =>
-    ObservableValue(initialValue, name: name);
+ObservableValue<T> observable<T>(T initialValue,
+        {String name, ReactiveContext context}) =>
+    ObservableValue(context ?? mobxContext, initialValue, name: name);
 
 /// Creates a computed value with an optional [name].
 ///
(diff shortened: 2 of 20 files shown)


lib/src/api/context.dart
```diff
@@ -0,0 +1,4 @@
+import 'package:mobx/src/core/context.dart';
+
+/// Global Mobx context.
+final ReactiveContext mobxContext = ReactiveContext();
```
pavanpodila: How about calling it `globalContext` or `defaultContext`?

lib/src/api/observable.dart
```diff
@@ -47,15 +49,18 @@ ObservableValue<T> observable<T>(T initialValue, {String name}) =>
...
 /// Creates a simple Atom for tracking its usage in a reactive context. This is useful when
 /// you don't need the value but instead a way of knowing when it becomes active and inactive
 /// in a reaction.
 ///
 /// Use the [onObserved] and [onUnobserved] handlers to know when the atom is active and inactive
 /// respectively. Use a debug [name] to identify easily.
-Atom createAtom({String name, Function onObserved, Function onUnobserved}) =>
-    Atom(name ?? 'Atom@${ctx.nextId}',
-        onObserve: onObserved, onUnobserve: onUnobserved);
+Atom createAtom({String name, Function onObserved, Function onUnobserved}) {
```
pavanpodila: No context argument here? Any reason why we are picking `mobxContext` directly?

pavanpodila: Also the context should be a named argument

lib/src/api/reaction.dart
```diff
@@ -56,8 +62,9 @@ ReactionDisposer when(
   bool Function() predicate,
   void Function() effect, {
   String name,
+  ReactiveContext context,
 }) =>
-    createWhenReaction(predicate, effect, name: name);
+    createWhenReaction(context ?? mobxContext, predicate, effect, name: name);
```
pavanpodila: I think the context should go into the named argument list

pavanpodila: I think that will keep the signature consistent throughout

lib/src/core/action.dart
```diff
@@ -2,14 +2,14 @@ import 'package:mobx/src/core/context.dart';
 import 'package:mobx/src/core/derivation.dart';
 
 class Action {
-  Action(Function fn, {String name}) {
-    _fn = fn;
+  Action(this._context, this._fn, {String name})
+      : name = name ?? _context.name('Action');
```
pavanpodila: This is a good addition: using the `context.name()`👍

lib/src/core/action.dart
```diff
@@ -30,40 +30,15 @@ class Action {
...
   }
 
   void _endAction(Derivation prevDerivation) {
-    ctx
+    _context
       ..endBatch()
       ..untrackedEnd(prevDerivation);
   }
 }
-
```
pavanpodila: Why are we removing `runInAction()`?

test/observable_test.dart
```diff
@@ -23,15 +24,15 @@ void main() {
   });
 
   test('Raw observables', () {
-    final x = ObservableValue(1000);
+    final x = ObservableValue(mobxContext, 1000);
```
pavanpodila: We will also need some tests that will provide custom context and verify its being used correctly.

pavanpodila: I think that will apply to actions, observables, reactions

lib/src/api/observable.dart
katis: Yeah, missed this one.

lib/src/api/context.dart
katis: I thought about this a little more, and if we want to explore the Zone thing at some point, maybe "currentContext" or "activeContext" would be a decent name? That way we can replace it with a getter if needed.

lib/src/api/reaction.dart
katis: I figured that `createWhenReaction` is an internal helper function, and as such the context should be a named argument. I've had trouble before in forgetting to pass this kind of stuff around if it's optional everywhere, so I prefer that just the public API's have it as an optional named argument.

lib/src/api/context.dart
pavanpodila: Yup that sounds good

pavanpodila: let's go with `currentContext`

lib/src/api/reaction.dart
pavanpodila: ok that makes sense

lib/src/core/action.dart
katis: It is now in api/action.dart

pavanpodila:
@katis Also from the next PR, you can use this repo with a separate branch. I made you a collaborator. That way I could also chip in if needed.

katis:
@pavanpodila made the changes requested, added tests using Mockito for context.
I added support for dev sdks in the version range too, since Flutter uses a dev version and my installation uses Flutters SDK.

PR #15, opened 2019-01-06
Make Interceptable use composition

katis:
The Interceptable made too many things public and the code was a bit javascripty, so I decided to refactor it a bit. Tried to add all functions into the mixin, but then all the mixin methods would be public in ObservableValue. I think composition made it cleaner.

I'll do the same thing to Listenable if this looks ok to you @pavanpodila

### DIFF
--- a/test/intercept_test.dart
+++ b/test/intercept_test.dart
@@ -1,5 +1,10 @@
 import 'package:mobx/mobx.dart';
+import 'package:mobx/src/core/atom.dart';
+import 'package:mobx/src/interceptable.dart';
+import 'package:mockito/mockito.dart';
 import 'package:test/test.dart';
+
+import 'shared_mocks.dart';
 
 void main() {
   test('intercept', () {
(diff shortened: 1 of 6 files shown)


katis:
There was also a bug in interceptChange, untracked used the global context. There's now a test for it.

test/intercept_test.dart
```diff
@@ -1,6 +1,11 @@
...
+import 'package:mobx/src/core/atom.dart';
+import 'package:mobx/src/interceptable.dart';
+import 'package:mockito/mockito.dart';
 import 'package:test/test.dart';
 
+import 'shared_mocks.dart';
+
 void main() {
   test('intercept', () {
     final x = observable(10);
```
pavanpodila: I think we also need a test where the `MockContext` is passed into the observable, action, reaction API directly. This will test if the context flows through properly and used correctly as well. I think the entire public API is missing the custom context tests?? 

This will be a good blackbox test of the context flowing from the user-level API

katis: MockContext tests exist for `Interceptors`, `action()`, `computed()`, `observable()` and `reaction()`

pavanpodila: Yup you are right. I think they are only missing for `autorun` and `when`

katis: Added context tests for `autorun` and `when`

PR #16, opened 2019-01-06
Use composition for Listenable

katis:
Same kind of change I did for Interceptable


pavanpodila:
@katis looks like the coverage went down a bit because of the addition of `untracked` to action.dart. Could you please add a test around that?

Coverall reference: https://coveralls.io/builds/20927521/source?filename=lib%2Fsrc%2Fapi%2Faction.dart

pavanpodila:
> @katis looks like the coverage went down a bit because of the addition of `untracked` to action.dart. Could you please add a test around that?
> 
> Coverall reference: https://coveralls.io/builds/20927521/source?filename=lib%2Fsrc%2Fapi%2Faction.dart

nevermind, I added the test

PR #17, opened 2019-01-06
ObservableList<T>

pavanpodila:
We could have used `noSuchMethod()` but will require dart:mirrors. Since we want to support Flutter eventually (which does not have `dart:mirrors`), we have to go the `@override` route.

The solution is pretty closely modeled around the mobx `ObservableArray<T>`. There is a single `Atom` that is used for the `ObservableList`. For read-methods, we call `atom.reportObserved()` and for write methods we call `atom.reportChanged()`.


katis:
You could use [IterableMixin](https://api.dartlang.org/stable/2.1.0/dart-collection/IterableMixin-class.html) for the Iterable methods.

Also, you could implement an iterator that reads the ObservableList while iterating and return that from `get iterator`. That way it should report reads even when passing around transformed iterables.

katis:
There's also [ListMixin](https://api.dartlang.org/stable/2.1.0/dart-collection/ListMixin-class.html), but it warns to overload `add` and `addAll`.

pavanpodila:
I am not clear how the _mixin_ can help here. I do need to _override_ and fire the `reportObserved()` and `reportChanged()` hooks on the **atom**. This will require going through all the methods anyways. Am I missing something here ?

pavanpodila:
@katis is this the way to do it ?

pavanpodila:
@katis could you please review?

pavanpodila:
> I don't see the use of ListMixin or IterableMixin anywhere, but I guess it can be refactored in later

I tried with the ListMixin but it was giving me a bunch of errors with the generic types. Also in my case my list is just a compositional type that wraps an inner list. I may be doing it wrong though :-(

katis:
Well, better to have this in now, Mobx isn't that useful without ObservableList 😄 . I can maybe take a look at some point if I can get the mixin way to work.

pavanpodila:
Ya, I am going to pick up on ObservableMap<T> as well, soon. Need to the add the `observe`/` intercept` hooks for ObservableList<T>. I think we should start creating a 0.1 milestone soon. We are getting there sooner than I thought. Huge thanks to you @katis 👍

Issue #14, opened 2019-01-06
Flutter API


katis:
Last night I thought I'd try to use Mobx with Flutter and created a simple widget based on the design of mobx-react and Flutter's [ValueListenableBuilder](https://docs.flutter.io/flutter/widgets/ValueListenableBuilder-class.html):

```dart
import 'package:flutter/widgets.dart';
import 'package:mobx/mobx.dart';

void noOp() {}

class Observer extends StatefulWidget {
  const Observer({Key key, @required this.builder}) : super(key: key);

  final Widget Function(BuildContext) builder;

  @override
  State<StatefulWidget> createState() => _ObserverState();
}

class _ObserverState extends State<Observer> {
  bool _isBuildPending = false;

  Reaction _buildReaction;

  @override
  void initState() {
    super.initState();

    _buildReaction = Reaction(mobxContext, () {
      if (_isBuildPending) {
        return;
      }

      _isBuildPending = true;
      if (!mounted) {
        return;
      }

      setState(noOp);
    });
  }

  @override
  Widget build(BuildContext context) {
    _isBuildPending = false;
    Widget result;
    _buildReaction.track(() {
      result = widget.builder(context);
    });
    return result;
  }
}
```

usage:

```dart
class Greeter extends StatefulWidget {
  @override
  State<StatefulWidget> createState() => _GreeterState();
}

class _GreeterState extends State<Greeter> {
  final name = observable('Jane');

  @override
  Widget build(BuildContext context) => Observer(builder: (context) {
    return Text('Hello, ${name.value}');
  });
}
```

It does seem to work and could be a starting point for the `flutter_mobx` API. Probably the life cycle of widgets will bring all kinds of problems like in `mobx-react`.

pavanpodila:
I think we can simply use a `reaction` that internally calls the `build` method. That way anything that is invoked in the `builder` method is tracked. I don't think we should use the internal `track()` method directly.

This is certainly the direction to take!

pavanpodila:
Got it. We can also try to build a **custom scheduler** that will call the reaction only when there is a change. I think that will keep it clean and only rely on the public interface? I want to avoid using `track` directly unless we find that there is no other way.

We haven't exposed the scheduler yet from a `reaction`, which we will have to do first.

pavanpodila:
[...]

PR #19, opened 2019-01-06
Tracking changes between start and end

katis:
We could design a pretty elegant Flutter integration with `flutter_hooks`, but it requires a way to track changes between start and end with no ability to model them as a callback.

I tried to design an API that supports the use case and hopefully prevents worst misuses,
but this should almost never be used.

Also added `dart:meta` and marked the Tracking class with an `@experimental` annotation.

See #14 for further discussion


test/tracking.dart
```diff
@@ -0,0 +1,57 @@
+import 'package:mobx/mobx.dart';
+import 'package:mobx/src/api/context.dart';
+import 'package:mobx/src/core/tracking.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('Tracking reacts to changes to reactive values between begin and end',
```
pavanpodila: Can we also have a test with a autorun() to ensure the reactions are kicked off when the observables change outside

lib/src/core/tracking.dart
```diff
@@ -0,0 +1,40 @@
...
+import 'package:mobx/src/core/context.dart';
+import 'package:mobx/src/core/derivation.dart';
+import 'package:mobx/src/core/reaction.dart';
+
+/// Tracks changes that happen between [Tracking.begin] and [Tracking.end].
+///
+/// This should only be used in situations where it is not possible to
+/// track changes inside a callback function.
+@experimental
+class Tracking {
```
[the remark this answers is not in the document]
katis: I'll refactor context a bit, add most of the logic there, but I think this should be the public (and unsafe) API of doing tracking without callbacks. The callback API should be used for 99.9% use cases, even in Reaction & ComputedValue.

katis:
I changed the `Tracking` class into a `DerivationTracker`. I tried implementing the `useObserver` hook, and realized that the tracker must be usable multiple times.

The API is now:
```dart
class DerivationTracker {
    DerivationTracker(ReactiveContext context, Function() onInvalidate,
      {String name});

  void start();

  void end();
  
  void dispose();
}
```

@rrousselGit I implemented the hook like this in my experiment, does it look OK to you?

```dart
void useObserver({ReactiveContext context}) => Hook.use(context == null
    ? const _ObserverHook()
    : _ObserverHook.withContext(context));

class _ObserverHook extends Hook<void> {
  final ReactiveContext context;

  const _ObserverHook() : context = null;

  _ObserverHook.withContext(this.context) : super(keys: [context.hashCode]); // not sure about the keys

  @override
  _ObserverHookState createState() => _ObserverHookState();
}

class _ObserverHookState extends HookState<void, _ObserverHook> {
  DerivationTracker _tracker;

  @override
  void initHook() {
    super.initHook();
    _initTracker();
  }

  @override
  void didUpdateHook(_ObserverHook oldHook) {
    super.didUpdateHook(oldHook);
    _tracker?.dispose();
    _initTracker();
  }

  void _initTracker() {
    _tracker = DerivationTracker(hook.context ?? currentContext, _invalidate);
  }

  void _invalidate() => setState(noOp);

  static void noOp() {}

  @override
  void build(BuildContext context) => _tracker.start();

  @override
  void didBuild() {
    super.didBuild();
    _tracker.end();
  }

  @override
  void dispose() {
    _tracker.dispose();
    super.dispose();
  }
}
```

katis:
@pavanpodila could we merge this now and think about the internal refactoring of ComputedValue etc. later?

lib/src/core/reaction.dart
```diff
@@ -1,27 +1,39 @@
...
+/// Tracks changes that happen between [start] and [end].
+///
+/// This should only be used in situations where it is not possible to
+/// track changes inside a callback function.
+@experimental
+class DerivationTracker {
+  DerivationTracker(ReactiveContext context, Function() onInvalidate,
       {String name})
-      : super(context, onInvalidate, name: name);
+      : _reaction = Reaction(context, onInvalidate, name: name);
```
pavanpodila: Not clear why we need a `reaction` here for tracking? Could you please elaborate? Is it mostly to wrap the `onInvalidate` in some derivation?

rrousselGit:
@katis There are some minor changes we can do, but that overall should do the trick, although I do not think there's a need for `didUpdateHook` at all.

pavanpodila:
@rrousselGit Can I add you as a collaborator? I think you bring valuable expertise that we all can benefit from?

rrousselGit:
Sure, please do so! 😄

pavanpodila:
> Sure, please do so! 😄

You are now! Please confirm on the email link from github :-). Thanks!

PR #25, opened 2019-01-07
Merge core files into framework.dart

katis:
Simply concatenates files in `core` to `framework.dart` and fixes imports around the application.

I will do more reorganizing and cleaning, but I'll do this in parts this time.
After file reorganization is complete, I'll start making things private where needed.


katis:
The framework file is too big, and apparently some bigger projects do use the `part` pattern, so I think I'll do it that way instead.

PR #30, opened 2019-01-08
Monorepo

katis:
Moved `mobx` package files to mobx/, used `mono_repo` to generate travis configs.


rrousselGit:
I think we should test to generate a flutter configuration before merging it. As it may be something non-trivial to do using mono_repo

pavanpodila:
@katis can we try having a flutter configuration?

katis:
Unfortunately can't get mono_repo to work with a flutter package. I need to do some build scripts manually for now.

rrousselGit:
I have an example of a travis-ci for flutter here: https://github.com/rrousselGit/flutter_hooks/blob/master/.travis.yml

It's not documented, but you can have the following in mono_pkg.yaml:

```yaml
stages:
  - some_stage_name:
    - command: do whatever you want
```

It should be doable to copy paste the commands from the previous `.travis.yaml` and prefix them by `command:` for it to work with `mono_repo`

katis:
But I guess "monorepo pub get" and "monorepo pub upgrade" wont work?

rrousselGit:
Ah – Indeed, didn't think of that. 
I'll see if I can make a PR for it. It shouldn't be too difficult from my memory

rrousselGit:
https://github.com/dart-lang/mono_repo/pull/118

I made a PR to support `mono_repo pub get` for flutter projects.

katis:
I'll merge this, and then start with the `flutter_mobx` project. I think we'll get by for a while without `mono_repo pub`

PR #32, opened 2019-01-08
Increase coverage

pavanpodila:
I had to introduce `onError` argument to `Reaction` and the public API to support exception handling. This was a because a test was checking if an exception causes the reaction to get disposed correctly. That resulted in adding the `onError` argument to all the reaction types.

@katis please verify to merge


lib/src/core/context.dart
```diff
@@ -231,27 +237,34 @@ class ReactiveContext {
         return true;
 
       case DerivationState.possiblyStale:
-        return untracked(() {
-          for (final obs in derivation._observables) {
-            if (obs is ComputedValue) {
-              // Force a computation
+        final prevDerivation = untrackedStart();
```
katis: Didn't changing this from a callback just make the code more error prone, as you have to remember to call `untrackedEnd` before every return? Use try-finally at least

test/computed_test.dart
```diff
@@ -102,4 +102,21 @@ void main() {
     verify(context.nameFor('Computed'));
     verify(context.trackDerivation(value, fn));
   });
+
+  test('computed catches exception in evaluation', () {
+    var shouldThrow = true;
+
+    final x = computed(() {
```
katis: Computed values in mobx re-throw exceptions thrown in the body. Will this come later?

lib/src/core/context.dart
pavanpodila: Agree @katis. I brought it back to the callback version of `untracked`

test/computed_test.dart
pavanpodila: Yes. Currently we have something called the **Error Boundary** in the MobX parlance. This means errors are trapped locally to keep the MobX system consistent. We will introduce global option later to disable this. I think we are good to merge

pavanpodila: Please see the `disableErrorBoundaries` option in [this page](https://mobx.js.org/refguide/api.html)

pavanpodila:
Merging now

PR #34, opened 2019-01-09
flutter_mobx project

katis:
Created flutter_mobx and a vanilla Flutter Observer component.

Still need to check the mono_repo stuff on a non-Windows computer.


katis:
@pavanpodila I got this to build on Travis, it runs package get, analysis and tests. I just can't figure out how to add Coveralls.

Travis config now has 2 env configs that makes it run separate builds for the packages:

```yaml
env:
  - PACKAGE=mobx
  - PACKAGE=flutter_mobx FLUTTER=true
```

Now we would need to somehow send different Coveralls reports from these I guess?

katis:
Codecov seems to support the monorepo use case https://docs.codecov.io/docs/flags

pavanpodila:
Let's try **codecov**. I am not attached to coveralls, just that the `dart_coveralls` had an easy integration. If we can find such a package, we are good to go 👍

katis:
@pavanpodila I had to let codecov request access to mobxjs organization, not sure who has to accept it?

pavanpodila:
> @pavanpodila I had to let codecov request access to mobxjs organization, not sure who has to accept it?

I've asked @mweststrate, who is the owner to approve.

mweststrate:
Approved. Note that it needed access to private repositories as well, that
is a little strange? (not really a problem atm, since there are none)

Op ma 14 jan. 2019 om 03:30 schreef Pavan Podila <notifications@github.com>:

pavanpodila:
Thanks @mweststrate. I think coveralls had the same requirement ?

flutter_mobx/lib/flutter_mobx.dart
```diff
@@ -0,0 +1,57 @@
...
+class Observer extends StatefulWidget {
+  const Observer({@required this.builder, Key key, this.context})
+      : assert(builder != null),
+        super(key: key);
+
+  final ReactiveContext context;
+  final ObserverBuilder builder;
+
+  @visibleForTesting
+  DerivationTracker createDerivationTracker(Function() onInvalidate) =>
```
pavanpodila: @katis Wondering if we even need a **DerivationTracker**. Why not use the **Reaction** directly, now that we can expose the `{start, end}Tracking` as separate calls? 🤔

pavanpodila:
Loving the codecov report 🏆 😃

flutter_mobx/lib/flutter_mobx.dart
katis: Frankly, I really don't like using the start/endTracking methods anywhere. The `track(callback)` interface is a superior one from a safety point of view. Consider a function like this:

```dart
foo() async {
  tracker.startTracking();
  await doSomethingAsync();
  tracker.endTracking();
}
```

Here the tracking starts, `doSomethingAsync` is called, returning a Future and the method is suspended. Then the Dart runtime schedules something else to execute while waiting the Future to resolve, making mobx track all sorts of random things. You can't create a mess like this with the callback interface. This is why the only place to use it should be the `useObserver` hook...

I think I'll figure a way to expose `Reaction.track` properly and use that instead here, for the above reasons, but I'll do it in a separate PR.

katis:
@mweststrate I also sent a request for adding a better integration with codecov to the repo

mweststrate:
Approved!

Op ma 14 jan. 2019 om 17:42 schreef katis <notifications@github.com>:

PR #36, opened 2019-01-13
Add observe and intercept hooks for ObservableList<T>

pavanpodila:
Adding the `observe` and `intercept` hooks for `ObservableList<T>`. This also includes some refactoring of the `Listenable` and `Interceptable` to extract out the common pieces

PR #37, opened 2019-01-13
Unwrap values in ObservableList

katis:
I tested the ObservableList in my own project, and I noticed that I really don't want ObservableList to wrap the items in an ObservableValue. If I want that behavior I can easily make a `ObservableList<ObservableValue<T>>` explicitly.

At the same time I used ListMixin, and made Iterable/Iterator returning methods lazy, so that the reportObserved happens after you start iterating with a for-loop or whatever.

We can merge this after #36, but I needed this in my project anyways, so here it is.


katis:
Just noticed that in MobxJS the indexes themselves are observable, so this needs some rethinking.

mobx/test/observable_list_test.dart
```diff
@@ -14,37 +17,59 @@ void main() {
...
-        'map': (_) => _.map((x) => x),
         'isEmpty': (_) => _.isEmpty,
         'isNotEmpty': (_) => _.isNotEmpty,
         'iterator': (_) => _.iterator,
-        'reversed': (_) => _.reversed,
         'single': (_) => _ignoreException(() => _.single),
         'first': (_) => _ignoreException(() => _.first),
         'last': (_) => _ignoreException(() => _.last),
         'toSet': (_) => _.toSet(),
         'toList': (_) => _.toList(),
```
pavanpodila: Seem to have skipped a few methods like `reversed`, `cast`, etc. Intentional ?

katis: Their reports are now lazy, so calling `observableList.map(fn)` doesn't report yet, but iterating the iterable returned by `map` will, e.g. ` observableList.map(fn).forEach((_) {})`.

The `fires reportObserved() lazily on iterator returning methods` test group handles those.

katis:
I'll merge your changes to this branch when I have the time.

katis:
Says "This branch has no conflicts with the base branch" for me

mobx/lib/src/api/observable_list.dart
```diff
@@ -1,61 +1,72 @@
...
-  }
+  Listeners<ListChangeNotification<T>> _listenersField;
 
-  @override
-  T get last {
-    _atom.reportObserved();
-    return _list.last;
-  }
+  Listeners<ListChangeNotification<T>> get _listeners =>
+      _listenersField ??= Listeners(_context);
```
pavanpodila: 👍 on the lazy init

mobx/lib/src/api/observable_list.dart
```diff
@@ -263,19 +150,6 @@ class ObservableList<T>
     return _list.toSet();
```
pavanpodila: Are we discarding the `growable` argument ??

katis: yeah, will fix it

katis:
I resolved your comments, merging

PR #40, opened 2019-01-15
Value factories

katis:
Replaced renamed `ObservableValue<T>` to `Observable<T>`, `ComputedValue<T>` to `Computed<T>`.

Changed `observable()`, `computed()` and `createAtom` to factories in their respective classes.

I only touched classes I felt were "value-like". I also think that the renaming works well in Dart because of generics, since you can say "this is an observable string", or "this is a computed int", instead of "this is a string computed value".

We discussed this before, and I thought I'd implement it now, so there's no need for big renames after the first real release. Another motivation is, it would be nice to have `observable` and `computed` names free for the code generation package annotation names. `action` is still reserved, which would also be nice as an codegen annotation. Maybe we could make `action0`, `action1`, `action2`, ... overloads for action builders of different arities and leave `action` free.


mobx/lib/src/core/observable.dart
```diff
@@ -1,11 +1,28 @@
...
+  /// var x = Observable(10);
+  /// var message = Observable('hello');
+  ///
+  /// print('x = ${x.value}'); // read an Observable's value
+  /// ```
+  factory Observable(T initialValue, {String name, ReactiveContext context}) =>
+      Observable._(context ?? mainContext, initialValue, name: name);
+
+  Observable._(ReactiveContext context, this._value, {String name})
       : _interceptors = Interceptors(context),
```
pavanpodila: May be we can init the **interceptors** and **listeners** lazily, just like you did for `ObservableList<T>` ?

katis: Let's do it in another PR

pavanpodila: agree..

PR #41, opened 2019-01-15
Use Reaction track instead of start/end tracking in Observer

katis:
I changed Observer to use `Reaction` track method instead of start/end tracking.

I ended up removing `DerivationTracker`, and using the methods in `Reaction` directly.
Dart conventions say that library consumers should not import from `package:<pkg>/src`, as they are internal details, so I made `Reaction` a public interface which currently has only a few things like `errorValue` and `dispose()`. We can add more stuff to it when we figure out the things we want to be in the public API. The internal implementation is `ReactionImpl`, which has `track` and `start/endTracking()`. `ReactionImpl` is not exported from `lib/mobx.dart`, but can be accessed from our Flutter packages by importing from `/src`.

I also renamed the `ReactionDisposer.$mobx` to just `reaction`, seemed like a javascripty name.


mobx/lib/src/api/reaction.dart
```diff
@@ -24,11 +24,11 @@ import 'package:mobx/src/core.dart';
 /// x.value = 30; // Will not cause autorun() to re-trigger as it's disposed.
 /// ```
 
-ReactionDisposer autorun(Function(Reaction) fn,
+ReactionDisposer autorun(Function(ReactionImpl) fn,
```
pavanpodila: I am not in favor of exposing `ReactionImpl` in the public signatures. Ideally the `Reaction` (the interface) is what we are exposing, whose implementation happens to be `ReactionImpl`. I think we should keep the interface here in the signature, like it was before:

`ReactionDisposer autorun(Function(Reaction) fn)`

pavanpodila: Same goes for all the other signatures. Since `ReactionImpl` implements **Reaction**, we should code to the interface rather than the implementation.

katis:
@pavanpodila ReactionImpl should now be gone from the public API

PR #42, opened 2019-01-16
fixing a few other calls that use ReactionImpl instead of Reaction

pavanpodila:


mobx/lib/src/core/context.dart
```diff
@@ -6,7 +6,7 @@ class _ReactiveState {
   int nextIdCounter = 0;
 
   Derivation trackingDerivation;
-  List<ReactionImpl> pendingReactions = [];
+  List<Reaction> pendingReactions = [];
```
katis: does pendingReactions need to be public?

pavanpodila: Its part of the `ReactiveState`, which is already private. So technically the whole state is private to the context.

pavanpodila: merging for now

PR #43, opened 2019-01-16
Flutter mobx hooks package

katis:
Created the package and implemented the hook based on the discussed design.
Needs more tests, so not done yet.


rrousselGit:
It would be interesting to see how the tracker with nested observers and with exceptions. To make sure everything works as expected

pavanpodila:
> It would be interesting to see how the tracker with nested observers and with exceptions. To make sure everything works as expected

Good point. I think the **onError** handler of the `reaction` should take care of that and report on the Flutter UI in some way (possibly with some overlay like in React Native).

rrousselGit:
`didBuild` of hooks is called even if there's a crash in `build` method. So there's no need for `onError`. But it's to make sure that `didBuild` works as intended, as I'm not 100% certain about it

katis:
Got sidetracked with codegen experimentation for a bit, continuing with this in the weekend

katis:
@rrousselGit It seems to me that the `didBuild` lifecycle doesn't work correctly for the Mobx use case.

I tried rendering nested widgets, and this is the call order I get:

```
START BUILD 1
START TRACKING 1
END BUILD 1

START BUILD 2
START TRACKING 2
END BUILD 2

END TRACKING 2
END TRACKING 1
```

This means that the parent widget ends up tracking changes inside the child widget. `didBuild` should be called right after the widgets `build` is called.

The right order would be:

```
START BUILD 1
START TRACKING 1
END BUILD 1
END TRACKING 1

START BUILD 2
START TRACKING 2
END BUILD 2
END TRACKING 2
```

rrousselGit:
Hum I can likely change that.

I'll try to fix that as soon as possible, but I won't be able to do anything before Tuesday.

rrousselGit:
Okey I got a fix but this blocks another use-case (calling setState after build), so I'm not satisfied with it.


By looking into Flutter sources I see a potential solution to have both use-cases, but it requires to make a PR. 
I'll keep you updated.

In the meantime, is there anything we can progress on other than `useObserver`?

rrousselGit:
As a side-note, I've just realized that`useObserver` is a lot more useful than just syntaxically

UseObserver could allow creating a custom hook that both obtain the store from an inherited widget and listen to it:

```dart
SomeStore useSomeStore() {
  useObserver();
  return SomeStore._of(useContext());
}
```

By doing so, it becomes impossible to obtain the store without listening to the values used.

This prevents developers from forgetting to observe it.

rrousselGit:
I'll actively work on this

`useObserver` seems to be fixed with my latest changes.  @katis can you confirm this?

There's a raw `useAutoRun`/`useWhen`/`useReaction` available (there will definitely be some changes. Right now they are disposed of too often). 

Are there any other candidates for hooks?

katis:
@rrousselGit yeah, the call order now seems right. Nice work.

`useComputed` might be useful for having some ad-hoc transformations?

Also, `_ObserverHook` doesn't need to be private since it's inside `src` and not exported from `flutter_mobx_hooks.dart`.

pavanpodila:
Just a minor nitpick... It should be _**useAutorun**_ instead of the capitalized ~~useAutoRun~~.

rrousselGit:
I've been messing with testing a bit. 

Do you know of a better way to test that the reaction is properly started/stopped other than by subclassing the hook? 

I really dislike having to subclass the hook class. It's really just an implementation detail and in theory, we should be able to write the hook as a function.

pavanpodila:
> I've been thinking:
> 
> We could allow overriding `mainContext` during testing, similarly to how Flutter allows overriding `defaultTargetPlatform`.
> 
> If we do so, it should then be possible to replace the `mainContext` with a `Mock` from mockito, which _should_ allow testing reactions without having to edit the Hook class

We already do that with a `MockContext` that extends `Mock` from `mockito` :-)

katis:
@rrousselGit I've been thinking for a while that ReactiveContext should be a factory for Reaction, Atom, etc. so that mocking it in tests could be used to test internal behavior easily.

rrousselGit:
@katis That's what I did with the implementation of `useAutorun`.

The only thing that bothers me is we're manually using `mainContext` like so:

```dart
  Reaction createReaction() {
    final ctx = hook.context ?? mainContext;
    final name = ctx.nameFor('ObserverHook-Reaction');
    return ReactionImpl(ctx, onInvalidate, name: name);
  }
```


We're calling `nameFor` (why btw?) but don't have any way to check that we do.

pavanpodila:
> @rrousselGit I've been thinking for a while that ReactiveContext should be a factory for Reaction, Atom, etc. so that mocking it in tests could be used to test internal behavior easily.

@katis that is a good idea! If that happens the default `autorun`/`reaction`/`when`/`action`/etc are just a reference to `ReactiveContext.main.autorun{reaction, when, action}`. 

Note that `ReactiveContext.main` will become the ~~`mainContext`~~ and it will no longer be exposed. 

What do you think?

rrousselGit:
I don't think that's a dart way of doing things.

We could just make `mainContext` a getter

```dart
final _mainContext = ReactiveContext();
ReactiveContext mainContextOverride;

ReactiveContext get mainContext {
  var context = _mainContext;
  assert(() {
    if (mainContextOverride) {
     context = mainContextOverride;
    }
    return true;
  }();
  return context;
};
```

pavanpodila:
> In that case the `Zone` is the right place, and it already allows overriding it

Ya that sounds like it. I think @katis is already exploring that option.

katis:
Yeah, my idea originally was that `mainContext` (then called `currentContext`) would be a getter that would get the context from the current Zone. For some reason this did not work properly with the unit test library, and we had more pressing issues, so I stopped working on it.

katis:
> @katis that is a good idea! If that happens the default `autorun`/`reaction`/`when`/`action`/etc are just a reference to `ReactiveContext.main.autorun{reaction, when, action}`.
> 
> Note that `ReactiveContext.main` will become the `mainContext` and it will no longer be exposed.
> 
> What do you think?

I think that we should keep the global variable, and keep the ReactiveContext class as something the users never have to refer directly (except maybe in unit tests or in advanced use).
But yes, I think that making the context a factory has merit, I'm just not sure if _all_ things should be constructed there, or only the lower level stuff (Atom, Reaction, ...)?

Also maybe we should change the `mainContext` back to `currentContext` before release, in case we want to use the Zone approach?

rrousselGit:
I still don't understand why we'd want to put all constructors for Reaction/whatever on Reactive Context

katis:
Because ReactiveContext is injected in pretty much everywhere, we could use it to create mock instances of Action/Reaction in tests. For example, at the moment ObservableHookState creates a ReactionImpl internally, so we can't mock and can't track the methods that are called by it.

rrousselGit:
> For example, at the moment ObservableHookState creates a ReactionImpl internally


The current implementation of `AutorunHook` is able to mock `autorun` function, without modifying the context.


While I get that an overridable method would reduce the boilerplate, this is definitely doable

rrousselGit:
Should we merge this branch and implements each hook separately? This PR is starting to get quite big.

katis:
I think we can merge this when the build is fixed

rrousselGit:
Is it fine to change the ci to use master branch for flutter then? 

Because hooks rely on that branch, as I made a few PR recently

katis:
Hmm... would it be possible to use the master only for the `flutter_mobx_hooks` package?

rrousselGit:
I tweaked the CI a bit so that it runs on both master and beta changes for flutter packages.
And `flutter_mobx_hooks` is set as allowed to fail on beta branch

I'm not a Travis expect though, so there's likely a way to optimize it.
Currently, it's pretty slow, as we're effectively downloading the dart SDK 9 times and flutter 6 times

From what I understand, I don't think we can improve perfs without a `language: flutter` support on travis.

katis:
Yeah, the slowness of Travis is already becoming a problem. Maybe next step is to combine the builds into a single run (or two for flutter/pure dart packages)

rrousselGit:
Maybe we should enable `auto cancel branch build`. Currently, we're running two builds simultaneously for each commit on a PR. That would free up some resources

katis:
I'm not sure, but I think codecov might be using the two builds to make the coverage diff?

rrousselGit:
It does not. These two builds are identical anyway, they point to the same sources.

rrousselGit:
Interestingly, running `pub run test` on `mobx` shows about 55 failing tests

Never used `test_coverage` before, is it truly trustworthy?

katis:
@rrousselGit I think it might be the `enforceActions` change. I guess both me and pavan run `pub run test test/all.dart` instead, which doesn't show that error

katis:
@rrousselGit I'll fix the test runs in a separate pr

rrousselGit:
Test passing with a 100% coverage.
À little review? :)

tool/test.sh
```diff
@@ -4,7 +4,7 @@ set -x
 
 if [[ "${FLUTTER}" = true ]]
 then
-  $FLUTTER_CMD test --coverage --coverage-path coverage/lcov.info test/all.dart
+  $FLUTTER_CMD test --coverage --coverage-path coverage/lcov.info
```
katis: Why was `all.dart` removed?

flutter_mobx_hooks/lib/src/use_when.dart
```diff
@@ -0,0 +1,84 @@
...
+class _WhenHookState extends HookState<void, WhenHook> {
+  mobx.ReactionDisposer disposer;
+
+  @override
+  void initHook() {
+    disposer = WhenHook.when(
+      hook.predicate,
+      null,
+      name: hook.name,
+      context: hook.context ?? mobx.mainContext,
```
katis: This hook doesn't rebuild if context changes

tool/test.sh
rrousselGit: `test_coverage` is not used on flutter packages. So all that this `all.dart` file did was simulating the default behavior of the command line.

On the other hand this `all.dart` causes issue if we add new test files and forget to add them in this file.

katis: `test_coverage` generates it's own combined file. I've used the `all.dart` to run all unit tests with a single Dart VM launch which usually makes them run faster than every file opening their own separate VM.

Issue #45, opened 2019-01-17
I'd like to contribute to mobx.dart with file generator


copydog:
Now, I am writing annotation generator for mobx.dart.
Here is the repo, https://github.com/copydog/mobx_annotation

If you are interested, let me know how I can contribute to the mobx.dart.

pavanpodila:
This is awesome 💯. I suggest you join our team as a collaborator and work on this monorepo. Once you become a collaborator, you can create a branch under this repo and add the **annotation** project at the root level, just like `mobx` or `flutter_mobx`. 

Let me know if that's ok with you and I'll add you 😃

copydog:
yes, it is ok with me, please add me to collaborator

pavanpodila:
Just sent you the invite! Once you accept, you should be able to create a branch and send a PR with the **annotations** project added as a peer-folder of `mobx`. Look forward to seeing it! 

I am sure it will drastically improve the the developer experience of using mobx.dart 👍

copydog:
Should I use the name 'mobx_annotation' or any other name for this project?

rrousselGit:
Code-generators should be published as two separate packages anyway (see https://github.com/google/built_value.dart, https://github.com/dart-lang/json_serializable)

As such we'll likely need two separate names.

katis:
I'd like the annotations to just live in the `mobx` package, that way you just need `mobx` and `mobx_annotations` in your pubspec. I made my own codegen experiments in #46, and you can see the project structure there.

pavanpodila:
The base annotations should be for `observable`, `computed`, `action`, which can live in the `mobx` package itself (as @katis suggested). From an end-user's perspective, it should come with _batteries-included_.

Similarly, we would also need the **`observer`** annotation for `flutter_mobx`.

copydog:
Just like mobx.js includes `observable`, `computed`, `action` in the main package, I think these annotations should be included in the `mobx` package (as @katis  @pavanpodila suggested). Also, `observer` annotation should be in `flutter_mobx` package.

Btw, I like Observer personally, because I want fully control to widgets that need rebuild. I think 'observer' annotation might not be useful for other user too.

Last thing, I vote the name 'mobx_annotations' (as @katis suggested), just like other *mobx.js* package `mobx-persist`, `mobx-devtools`, etc.

katis:
[...]

PR #46, opened 2019-01-17
Code generator experimentation

katis:
I also happened to experiment with source_gen and code generation (just noticed #45).

Very quick&dirty implementations for `@observable` and `@computed`. I used Atoms instead of
Observables for fields, as the field get's created in the superclass anyways, and using it makes the initialization simple. Also, listening (with `observe(listener)`) the Observable is not possible because it's field would be hidden anyway.

The `mobx_codegen` project contains the `source_gen` generator.

The `example` project has an example of configuring it to a project and running `pub run build_runner build` will re-run code generation.

The generator creates an inherited class that overrides annotated fields/methods and makes them reactive:

```dart
// generator_example.dart

part 'generator_example.g.dart';

@observable
abstract class User {
  User._();
  factory User() = _$User;

  @observable
  String firstName = 'Jane';

  @observable
  String lastName = 'Doe';

  @computed
  String get fullName => '$firstName $lastName';
}
```

=>

```dart
// GENERATED CODE - DO NOT MODIFY BY HAND

part of 'generator_example.dart';

// **************************************************************************
// ObservableGenerator
// **************************************************************************

class _$User extends User {
  _$User() : super._() {
    _$fullNameComputed = Computed(() => super.fullName);
  }

  final _$firstNameAtom = Atom(name: 'User.firstName');

  @override
  String get firstName {
    _$firstNameAtom.reportObserved();
    return super.firstName;
  }

  @override
  set firstName(String value) {
    super.firstName = value;
    _$firstNameAtom.reportChanged();
  }

  final _$lastNameAtom = Atom(name: 'User.lastName');

  @override
  String get lastName {
    _$lastNameAtom.reportObserved();
    return super.lastName;
  }

  @override
  set lastName(String value) {
    super.lastName = value;
    _$lastNameAtom.reportChanged();
  }

  Computed<String> _$fullNameComputed;

  @override
  String get fullName => _$fullNameComputed.value;
}
```


copydog:
I noticed you create an `Atom` for each user-defined field. (I know this is an experiment)
On future release, we should create an `Observable` for each field, because `Observable` can check new value automatically. Am I correct?

katis:
> 
> 
> I noticed you create an `Atom` for each user-defined field. (I know this is an experiment)
> On future release, we should create an `Observable` for each field, because `Observable` can check new value automatically. Am I correct?

I think that `Atom` might be the better way to go for `@observable`. Writing `@observable String foo = 'foo';` in the store class means that a field with storage is created for the `foo` variable even if you override it in the generated class. If we then create a hidden field `Observable<T>` and store the value inside it, the `super.foo` field goes to waste. It might then need to be nulled to clear the memory of the initial value, and the `Observable` value should be initialized from it's value, making the code a bit more complex.

`Observable` is pretty much an `Atom` + field for the value + event listeners. Since the class already has a field, and we can't use event listeners if the `Observable` value is hidden, using an Atom makes sense.

pavanpodila:
I think there should also be a provision to get the underlying atom for observation, say `$field`. This will be needed if someone wanted to attach `onObserve`, `onUnobserve` handlers.

copydog:
@pavanpodila Good point, I spend half day to think over it. I think we should keep generated `Observable` or `Atom` public-accessible by using `$field` instead of `_field`.

In order to make it self-accessible, we should generate a parent class not a sub-class (a mixin can not have a constructor that we need), then use it like `User extends _$User`

katis:
@pavanpodila this PR is a pretty good starting point for the code generator, it has a separate package for it, build runner configuration and an example that uses code generation. Could we just merge this and then we'll have a base to build on?

pavanpodila:
There is a drop in code coverage. Will to fix that before I can merge

katis:
> 
> 
> Looking great🥇. I would love to see `@action` implemented as well in the next PR 🚀

I was actually working on `@action`, it's now available in this branch, but the `ActionController`, `start/end` model needs to be discussed.

I'll get to work on some tests, I don't have any idea how to test code generators though. Might have a look how `built_value` does it's tests.

mobx/lib/src/api/action.dart
```diff
@@ -33,7 +33,7 @@ import 'package:mobx/src/core.dart';
 /// notifications and propagates them only after the completion of the action. Actions
 /// can also be nested inside, in which case the change notification will propagate when
 /// the top-level action completes.
-Action action(Function fn, {String name, ReactiveContext context}) =>
+Action createAction(Function fn, {String name, ReactiveContext context}) =>
```
pavanpodila: Why not just expose `Action()` directly and handle the null context inside? That way we don't need a `createAction`

katis: Added the factory for Action

katis:
@pavanpodila can we merge this now?
The generator will change a lot, and the only other coverage change seems to be kind of false positive from the private constructors of annotations.

mobx/README.md
```diff
@@ -13,56 +12,52 @@
...
 
-
 ## Cross cutting features
 
-- [x] Observability API for `ObservableValue` and `ComputedValue` 
-    - [x] `observe`
-    - [x] `intercept`
-    - [x] `onBecomeObserved`
-    - [x] `onBecomeUnobserved`
+- [x] Observability API for `ObservableValue` and `ComputedValue`
```
pavanpodila: Have to rename here as well `ObservableValue` => `Observable`

katis: Renamed

katis:
@pavanpodila renames are done

pavanpodila:
@katis there is a drop in test coverage but we can address that in a separate PR. I had set the branch rules to require not to have a drop in coverage. I have now relaxed that rule. Merging for now.

PR #47, opened 2019-01-18
Example todos

pavanpodila:
Really simple example for incrementing counter using an `Observable` and `Observer`.


mobx/example/todos/lib/todos.dart
```diff
@@ -0,0 +1,87 @@
...
+}
+
+enum VisibilityFilter { all, pending, completed }
+
+class TodoList {
+  TodoList() {
+    _visibleTodos = Computed(() {
+      switch (filter) {
+        case VisibilityFilter.pending:
+          return pendingTodos;
```
katis: Could pending/completed todos just be calculated here? No need for extra Computed fields

mobx/example/todos/lib/main.dart
```diff
@@ -0,0 +1,65 @@
...
+  Observable<int> counter = Observable(0);
+  Action incrementCounter;
+
+  @override
+  Widget build(BuildContext context) => Scaffold(
+        appBar: AppBar(
+          title: Text(widget.title),
+        ),
+        body: Center(
+          child: Observer(
```
katis: You could move the Observer to be the parent of the Text widget that reads `counter.value` to optimize the update scope a bit.

mobx/example/todos/lib/todos.dart
pavanpodila: I kept it separate in order to leverage caching of the computed property. Else we would be recomputing every time we request `pendingTodos`

pavanpodila:
Any idea why this is failing: https://travis-ci.com/mobxjs/mobx.dart/jobs/171335894#L717

Should I be putting this example under a different `library` directive?

katis:
The example should probably be in `flutter_mobx`, because the `mobx` package is built and tested with the regular Dart SDK, not Flutter SDK

pavanpodila:
I think the problem is in the `get.sh` script that is running at the top level and expecting a `pubspec.yaml` under example. Can we do something to make sure the packages get command only executes for the `$PACKAGE` and `$PACKAGE/test` directories ?

katis:
Commands are configured in `.travis.yml` only for the dirs configured here:

```
env:
  - PACKAGE=mobx
  - PACKAGE=flutter_mobx FLUTTER=true
```

I'm unfamiliar with how pub/flutter builds or tests example projects inside the main project.

Issue #48, opened 2019-01-19
Code generator design


katis:
Since we've started working on the code generation, I thought I'd open a discussion thread about its design.

We need at least three working annotations, `@observable`, `@computed` and `@action`.

The generator will find all classes with a specific annotation (`@observable`, `@store`?) and creates a private subclass `_$ParentClass` that overrides the annotated fields, getters and methods with Mobx magic.

I'll use parts of this annotated class in the next sections:

```dart
@store // tbd
class User {
  @observable
  String firstName = 'Jane';

  @observable
  String lastName = 'Doe';

  @computed
  String get fullName => '$firstName $lastName';

  @action
  void updateNames({String firstName, String lastName }) {
    ...
  }
}
```

## `@observable`

The first approach I tried was with a hidden `Observable<T>` field:

```dart
class _$User {
  final _$firstName = Observable<String>(null, name: 'User.firstName');

  @override
  String get firstName => _$firstName.value;

  @override
  set firstName(String value) => _$firstName.value = value;
}
```

This had some downsides. One was that the `User` creates storage for the original `firstName` field which then ends up unused. Second is that the initialization becomes slightly more complicated. The initial value must be copied from the superclasses field, and the field must be nulled so that it doesn't keep the initial value in memory. Code would look something like this:

```dart
class _$User {
 _$User() {
   _$firstName.value = super.firstName;
   super.firstName = null;
 }

 final _$firstName = Observable<String>(null, name: 'User.firstName');

  @override
  String get firstName => _$firstName.value;

  @override
  set firstName(String value) => _$firstName.value = value;
}
```

My second attempt was with an Atom which made the initialization a non-issue, since it just uses the original field:

```dart
class _$User {
 final _$firstNameAtom = Atom(name: 'User.firstName');

  @override
  String get firstName {
     _$firstNameAtom.reportObserved();
    return super.firstName;
  }

  @override
  set firstName(String value) {
    super.firstName = value;
    _$firstNameAtom.reportChanged();
  }
}
```

I think the Atom method is better, for the reasons above.

## `@computed`

Computed should be simple, just override a getter and use the original one:

```dart
class _$User {
  _$User() {
    _$fullNameComputed = Computed(() => super.fullName);
  }

 Computed<String> _$fullNameComputed;

  @override
  String get fullName => _$fullNameComputed.value;
}
```

## `@action`

[...]

PR #49, opened 2019-01-19
Use `implements Store` for finding classes that require code generation

katis:


katis:
Let's not merge this yet, I'll try a mixin based approach first as discussed in #48

katis:
Rolled into #51

PR #50, opened 2019-01-19
Codegen template refactor

katis:
Moved the templating in `src/templates`.

I tried mustache templates but my clumsy fingers couldn't handle not having autocomplete and types. I made my own template classes using Dart template strings and overriding toString. If you really want mustaches someone else can do them 😄.

Built on #49


katis:
Rolled into #51

PR #52, opened 2019-01-20
More examples

pavanpodila:
Trying with the counter and todos example. Todos is failing right now. @katis need your inputs

### DIFF
--- a/flutter_mobx/example/lib/counter.dart
+++ b/flutter_mobx/example/lib/counter.dart
@@ -2,6 +2,22 @@ import 'package:flutter/material.dart';
 import 'package:flutter_mobx/flutter_mobx.dart';
 import 'package:mobx/mobx.dart';
 
+part 'counter.g.dart';
+
+@observable
+class Counter {
+  factory Counter() = _$Counter;
+  Counter._();
+
+  @observable
+  int value = 0;
+
+  @action
+  void increment() {
+    value++;
+  }
+}
+
 class CounterExample extends StatefulWidget {
   const CounterExample({Key key, this.title}) : super(key: key);
 
@@ -12,18 +28,7 @@ class CounterExample extends StatefulWidget {
 }
 
 class _CounterExampleState extends State<CounterExample> {
-  _CounterExampleState() {
-    incrementCounter = action(() {
-      counter++;
-    });
-  }
-
-  final Observable<int> _counter = Observable(0);
-
-  int get counter => _counter.value;
-  set counter(int value) => _counter.value = value;
-
-  Action incrementCounter;
+  final _counter = Counter();
 
   @override
   Widget build(BuildContext context) => Scaffold(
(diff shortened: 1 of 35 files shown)


pavanpodila:
It fails with the following error: 

```
error: '_$TodoList.pendingTodos' ('() → ObservableList<dynamic>') isn't a valid override of 'TodoListBase.pendingTodos' ('() → ObservableList<Todo>'). (invalid_override at [flutter_mobx_example] lib/todos.dart:18)
```

Essentially the generic return types are not treated correctly.

katis:
@pavanpodila I'll take a look, my guess is that just getting a type name in codegen doesn't get the generic arguments.

pavanpodila:
Yup, that's the issue. We need a deeper introspection to construct the Type name. Not sure if Element supports that case but wasn't obvious from the API docs.

katis:
Yeah, I think I know how to fix it. Can I push to this branch in a moment?

pavanpodila:
Sure @katis

katis:
@pavanpodila I used element `displayName` instead of `name` and that seems to work.

flutter_mobx/example/lib/counter.dart
```diff
@@ -2,6 +2,21 @@ import 'package:flutter/material.dart';
 import 'package:flutter_mobx/flutter_mobx.dart';
 import 'package:mobx/mobx.dart';
 
+part 'counter.g.dart';
+
+class Counter = CounterBase with _$Counter;
+
+@observable
```
amrnt: You don't need this decoration anymore as the class implements `Store`

pavanpodila:
@katis the `displayName` fix seems to work for `@observable` annotations but still creating computed props without the generic type.

For example:
```dart
  @computed
  ObservableList<Todo> get pendingTodos =>
      todos.where((todo) => todo.done != true);
```
is still creating:

```dart
  @override
  ObservableList get pendingTodos => (_$pendingTodosComputed ??=
          Computed<ObservableList>(() => super.pendingTodos))
      .value;
  Computed<ObservableList> _$completedTodosComputed;
```

Note the loss of the type arguments for `ObservableList`

katis:
Worked for me, just checked out the code and re-ran the codegen.

flutter_mobx/pubspec.yaml
```diff
@@ -1,24 +1,24 @@
...
+description: Flutter Mobx.dart integration.
+version: 0.0.1
+author:
+homepage:
+
+environment:
+  sdk: '>=2.1.0-dev <3.0.0'
+
+dependencies:
+  mobx: ^0.0.19
```
katis: use the incremented 0.0.20 version

flutter_mobx/example/pubspec.yaml
```diff
@@ -20,10 +20,18 @@ dependencies:
   # The following adds the Cupertino Icons font to your application.
   # Use with the CupertinoIcons class for iOS style icons.
   cupertino_icons: ^0.1.2
+  mobx: ^0.0.19
```
katis: use the incremented 0.0.20 version

pavanpodila: worked after running clean on the `build_runner`

` flutter packages pub run build_runner clean`

pavanpodila:
@katis please review and merge if it looks good

PR #53, opened 2019-01-20
Relax codegen mixin rules

katis:
Allows you to write codegen like this (the `class Admin = AdminBase with _$Admin` syntax still works):

```dart
class Admin extends AdminBase with _$Admin {
  Admin(int id) : super(id);
}

abstract class AdminBase extends User implements Store {
  AdminBase(int id) : super(id);

  @observable
  String userName = 'admin';

  @observable
  ObservableList<String> accessRights = ObservableList();
}
```

Makes you reimplement base class constructor signature, but should work better with `json_serializable`

### DIFF
--- a/example/lib/src/generator_example.dart
+++ b/example/lib/src/generator_example.dart
@@ -25,7 +25,9 @@ abstract class UserBase implements Store {
   }
 }
 
-class Admin = AdminBase with _$Admin;
+class Admin extends AdminBase with _$Admin {
+  Admin(int id) : super(id);
+}
 
 abstract class AdminBase extends User implements Store {
   AdminBase(int id) : super(id);
--- a/example/lib/src/generator_example.dart
+++ b/example/lib/src/generator_example.dart
@@ -32,4 +32,7 @@ abstract class AdminBase extends User implements Store {
 
   @observable
   String userName = 'admin';
+
+  @observable
+  ObservableList<String> accessRights = ObservableList();
 }
--- a/example/lib/src/generator_example.g.dart
+++ b/example/lib/src/generator_example.g.dart
@@ -68,4 +68,18 @@ mixin _$Admin on AdminBase, Store {
     super.userName = value;
     _$userNameAtom.reportChanged();
   }
+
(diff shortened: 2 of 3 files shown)


example/lib/src/generator_example.dart
```diff
@@ -25,11 +25,16 @@ abstract class UserBase implements Store {
   }
 }
 
-class Admin = AdminBase with _$Admin;
+class Admin extends AdminBase with _$Admin {
```
pavanpodila: It still seems like a lot of ceremony for the store classes. How about we take all the code in `User` and make it into a `UserBase` class and then have User extend from it?

```dart
class UserBase with _$User { /* All the code unwrapped from annotations */ }

class User extends UserBase { /* Contains clean getters/setters and action-methods */}
```

I want to avoid having the end user do anything special. Kind of following the **It Just Works™️** principle.

Note: I am just postulating here. Not sure if it can be pulled off easily. But I really want to eliminate the extra work. I think this will be a killer feature when we introduce mobx.dart 💯

rrousselGit: I think we should allow both `class Admin = AdminBase with _$Admin;` and this syntax to coexist. So that users who don't care about serialization can just use the shorthand, and those who do will take the hard route.

pavanpodila: Why don't we remove the mixin based approach and instead use composition? As in take all the code in the original User class add wrappers for them. 

I think the action methods are tricky but we have already have the solution in terms of `startAction` - `endAction`. We take the original code of `User`'s action methods and wrap them inside `start/endAction`. 

For **observables** we use atoms and for **computed**, we create them inside a constructor. If none is provided, we synthesize one. 

Hope that makes sense.

katis: Because the original class can call it's own methods/fields and those wouldn't have observability.

katis: I made a suggestion for a small fix in https://github.com/mobxjs/mobx.dart/issues/48#issuecomment-455967671

PR #54, opened 2019-01-22
Add checks and errors for codegen

katis:
Added some validations for the codegen.

final/static `@observable` fields, and static/async `@action` methods generate a warning in the console and prevent code generation.

Needs more in the future, but this is a start.

### DIFF
--- a/mobx_codegen/lib/src/errors.dart
+++ b/mobx_codegen/lib/src/errors.dart
@@ -1,0 +1,122 @@
+R Function(T) withIndex<R, T>(R Function(T, int) action, {int start = 0}) =>
+    (T value) => action(value, start++);
+
+abstract class CodegenError {
+  bool get hasErrors;
+  String get message;
+}
+
+class StoreClassCodegenErrors implements CodegenError {
+  StoreClassCodegenErrors(this.name) {
+    _errorCategories = [
+      staticObservables,
+      staticActions,
+      finalObservables,
+      asyncActions
+    ];
+  }
+
+  final String name;
+
+  final PropertyErrors finalObservables = FinalObservableFields();
+  final PropertyErrors staticObservables = StaticObservableFields();
+  final PropertyErrors staticActions = StaticActionMethods();
+  final PropertyErrors asyncActions = AsyncActionMethods();
+
+  List<CodegenError> _errorCategories;
+
+  String get message {
+    final errors = _errorCategories
+        .where((category) => category.hasErrors)
+        .map(withIndex((category, i) => '  $i. ${category.message}', start: 1))
+        .join('\n');
+
+    return 'Could not make class "$name" observable. Changes needed:\n$errors';
+  }
+
+  bool get hasErrors => _errorCategories.any((category) => category.hasErrors);
+}
+
+abstract class PropertyErrors implements CodegenError {
+  final NameList _properties = NameList();
+
+  bool addIf(bool condition, String propertyName) {
+    if (condition) {
+      _properties.add(propertyName);
+    }
+    return condition;
+  }
+
+  String get propertyList => _properties.toString();
+
+  Pluralize propertyPlural = const Pluralize('the field', 'fields');
+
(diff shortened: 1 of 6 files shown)


mobx_codegen/lib/src/errors.dart
```diff
@@ -0,0 +1,122 @@
...
+  @override
+  String get message => 'Remove final modifier from $property $propertyList';
+}
+
+class StaticObservableFields extends PropertyErrors {
+  @override
+  String get message => 'Remove static modifier from $property $propertyList';
+}
+
+class AsyncActionMethods extends PropertyErrors {
```
pavanpodila: It is quite valid and common to have async action methods. The sync statements will be part of the action and the ones that execute after the `await` must be wrapped in `runInAction`. You can probably print a warning to ensure that the code executing after the `await` is wrapped in `runInAction()`, especially if several observables will be changed.

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -59,69 +62,108 @@ class StoreMixinVisitor extends SimpleElementVisitor {
...
-        ..atomName = '_\$${element.name}Atom'
-        ..type = element.type.displayName
-        ..name = element.name;
-
-      _storeTemplate.observables.add(template);
+    if (!_observableChecker.hasAnnotationOfExact(element)) {
+      return null;
     }
+
+    if (_checkFieldErrors(element)) {
```
pavanpodila: I would recommend a name like `hasFieldErrors` or in general `hasXXXErrors` so it reads more like english ;-).

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -59,69 +62,108 @@ class StoreMixinVisitor extends SimpleElementVisitor {
...
-        ..positionalParams = positionalParams.map(param)
-        ..optionalParams = optionalParams.map(param)
-        ..namedParams = namedParams.map(param);
-
-      _storeTemplate.actions.add(template);
+    if (!_actionChecker.hasAnnotationOfExact(element)) {
+      return null;
+    }
+
+    if (_checkMethodErrors(element)) {
```
pavanpodila: `hasXXXErrors`

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -59,69 +62,108 @@ class StoreMixinVisitor extends SimpleElementVisitor {
...
+      ..positionalParams = positionalParams.map(param)
+      ..optionalParams = optionalParams.map(param)
+      ..namedParams = namedParams.map(param);
+
+    _storeTemplate.actions.add(template);
+
+    return null;
   }
+
+  bool _checkMethodErrors(MethodElement element) => any([
```
pavanpodila: `hasXXXErrors`

mobx_codegen/lib/src/errors.dart
katis: Can't analyze the code inside method bodies (besides maybe getting local variable declarations).

pavanpodila: I think we should skip this check for async actions, as its common to have it

katis: But it doesn't work... We do have to figure out the mobx.dart async story, maybe with something like [computed-async](https://github.com/danielearwicker/computed-async-mobx) or providing helpers like those in [mobx-utils](https://github.com/danielearwicker/computed-async-mobx), but I really don't want to allow @action for async methods and then try to explain users why it'll only work for the code before the first await.

pavanpodila: Perhaps this could be a lint rule rather than an error? I see your point about keeping it simple and making all actions sync only. Let's start with an error then and relax later when we have a better solution.

katis:
@pavanpodila changed the validation method names

PR #55, opened 2019-01-22
Example todos

pavanpodila:
Adding a longer example: Todo List with core and computed observables


flutter_mobx/example/lib/todos/todos.dart
```diff
@@ -0,0 +1,185 @@
...
+  @computed
+  ObservableList<Todo> get pendingTodos =>
+      ObservableList.of(todos.where((todo) => todo.done != true));
+
+  @computed
+  ObservableList<Todo> get completedTodos =>
+      ObservableList.of(todos.where((todo) => todo.done == true));
+
+  @computed
+  bool get hasCompletedTodos => completedTodos.isNotEmpty;
```
katis: Getters that just read a simple property probably don't need to be `@computed`, just adds extra overhead with no benefit.

flutter_mobx/example/lib/todos/todos.dart
```diff
@@ -0,0 +1,185 @@
...
+    todos.add(todo);
+    currentDescription = '';
+  }
+
+  @action
+  void removeTodo(Todo todo) {
+    todos.removeWhere((x) => x == todo);
+  }
+
+  // ignore: use_setters_to_change_properties
```
katis: Just disable this lint rule completely, goes against the spirit of actions.

flutter_mobx/example/lib/todos/todos.dart
```diff
@@ -0,0 +1,185 @@
...
+
+  @action
+  void removeCompleted() {
+    todos.removeWhere((todo) => todo.done);
+  }
+
+  @action
+  void markAllAsCompleted() {
+    // ignore: avoid_function_literals_in_foreach_calls
+    todos.forEach((todo) => todo.done = true);
```
katis: Use a for loop

flutter_mobx/example/lib/todos/todos.dart
```diff
@@ -0,0 +1,185 @@
...
+  State<StatefulWidget> createState() => _TodoExampleState();
+}
+
+class _TodoExampleState extends State<TodoExample> {
+  final _list = TodoList();
+
+  final _textController = TextEditingController(text: '');
+
+  @override
+  Widget build(BuildContext context) => Scaffold(
```
katis: Maybe split into smaller widgets? I can understand if you want to keep the example simple though

flutter_mobx/example/lib/todos/todos.dart
pavanpodila: Agree but this also adds the vocabulary for the domain. Rather than doing `completedTodos.isNotEmpty` every single time, there is a conceptual vocabulary that you can use which is equally reactive. Also want to highlight the fact that you can have computed properties which depend on other computeds

flutter_mobx/example/lib/todos/todos.dart
pavanpodila: Sorry, are you suggesting to disable the rule or NOT to disable? 🤔 😄

flutter_mobx/example/lib/todos/todos.dart
pavanpodila: Ya was going to do that next

flutter_mobx/example/lib/todos/todos.dart
katis: Disable it completely from the project (and others too)

pavanpodila: Also I added the action to indicate the semantic operation of changing a filter or setting a  description for the new todo. Unfortunately its a one-liner and failing the lint. I rather not have a stray observable change outside of an action. This is also a design philosophy in MobX and it guards it with a configurable option called `enforceActions`.

katis: Remove the ignore comments for the removed rule

PR #57, opened 2019-01-23
moving top level example under mobx_codegen

pavanpodila:
Moving the code generator example under `mobx_codegen`. I think that's the right place.


mobx_codegen/example/test/generator_example_test.dart
```diff
@@ -1,5 +1,5 @@
-import 'package:example/src/generator_example.dart';
 import 'package:mobx/mobx.dart';
+import '../lib/src/generator_example.dart';
```
katis: Should be `import 'package:mobx_codegen_example/src/generator_example.dart'`?

pavanpodila: It was failing with all these errors, if I did that:

```shell
 error • Target of URI doesn't exist: 'package:mobx_codegen_example/example.dart' at example/test/generator_example_test.dart:2:8 • uri_does_not_exist
  error • Undefined class 'User' at example/test/generator_example_test.dart:6:37 • undefined_class
  error • The function 'User' isn't defined at example/test/generator_example_test.dart:62:19 • undefined_function
  error • The function 'Admin' isn't defined at example/test/generator_example_test.dart:62:43 • undefined_function
  error • The function 'Admin' isn't defined at example/test/generator_example_test.dart:65:19 • undefined_function
  error • The function 'Admin' isn't defined at example/test/generator_example_test.dart:77:19 • undefined_function

```

katis: alright, leave it like that then :S

PR #58, opened 2019-01-23
Add ObservableMap

katis:
Added tests, combined ObservableList and ObservableMap into the same module so that list could use the hidden `_wrap` constructor of the map to implement `asMap()`.

Still missing `observe`


mobx/lib/src/api/observable_collections.dart
```diff
@@ -0,0 +1,9 @@
+import 'dart:collection';
+import 'dart:math';
+
+import 'package:meta/meta.dart';
+import 'package:mobx/mobx.dart';
+import 'package:mobx/src/core.dart';
+
+part 'observable_collections/observable_list.dart';
```
pavanpodila: like the naming (`observable_collections`)

pavanpodila:
coverage needs to be improved though

katis:
Coverage back to positive, merging

katis:
I hadn't merged this for some reason 😕

pavanpodila:
Merging for now...we can add the `observe` in the next PR

PR #81, opened 2019-02-03
Don't allow FutureOr in future methods

katis:
Remove FutureOr support in `@observable` async methods. Doesn't really work right now, and wrapping FutureOr with ObservableFuture would remove the advantage of FutureOr anyways.


katis:
ObservableStream has a better fix.

PR #84, opened 2019-02-03
Subscribe/pause stream when (un)observed

katis:
Lazily subscribes to the stream when observing data/value/error/status. Pauses the stream when unobserved.

Stream combinators (map/filter/etc.) now return an ObservableStream.
I made the observables/actions a separate object that is initialized on first observation, so that using combinators should be cheap. Will look into making ObservableFuture do the same kind of initialization.

Also changed the StreamStatus enum to use the same names as ConnectionStatus in Flutter.


pavanpodila:
tests are taking a hit

Issue #87, opened 2019-02-04
Add enforceActions support in ReactiveConfig and Derivations


pavanpodila:
Support the [`enforceActions`](https://github.com/mobxjs/mobx/blob/gh-pages/docs/refguide/api.md#enforceactions) feature of MobX JS

katis:
I know that in Mobx JS enforced actions are disabled by default but I'd suggest that we enabled them by default, especially if we can get async `@actions` to work. This way making actions to mutate state would be the norm, and making state changes everywhere willy-nilly the "quick&dirty" way.

pavanpodila:
Agree @katis, let me try that in the next PR. Have a bunch of tests that need to be changed as we are mutating lot of the observables outside actions

katis:
@pavanpodila What about using `observed` as the default? That's what I use in all my JS projects.

pavanpodila:
Ya, that would be good. Even there about **62** tests are failing

pavanpodila:
I can take it up once we merge this. Please review :-)

katis:
If `enforceActions` were changed to be variable in the context, you could just disable enforcement in the test files.

pavanpodila:
Fixed in #88. Will open a separate issue for a stricter enforcement

PR #88, opened 2019-02-05
Enforce actions

pavanpodila:
- Checks if an observable is being mutated inside a computed function
- Will be adding support for `enforceActions` (#87)


analysis_options.yaml
```diff
@@ -71,7 +71,6 @@ linter:
     - parameter_assignments
     - prefer_adjacent_string_concatenation
     - prefer_asserts_in_initializer_lists
-    - prefer_bool_in_asserts
```
pavanpodila: this was being warned as deprecated

PR #90, opened 2019-02-06
Add tests for stream combinators

katis:
More to come, I need to do some refactoring on ObservableFuture first.


katis:
@pavanpodila can you review this

PR #93, opened 2019-02-06
setting the default enforceActions = EnforceActions.observed

pavanpodila:


pavanpodila:
@katis any idea why `mobx_codegen` is failing on tests?

katis:
@pavanpodila codegen example tests was missing enfore config

pavanpodila:
> @pavanpodila codegen example tests was missing enfore config

Aha. Wasn't clear from the error message on the console! Nice fix :-)

PR #111, opened 2019-02-12
Check observer

pavanpodila:
This will check the builder function and assert if no observables are detected. I ran into this issue a few days back and took quite some time to figure out that there were no observables being tracked inside! I think this is a good sanity check to have.


kholdunn:
can we make this optional?
In my case, I want the class to be **observer** all the time.

Thanks in advance

pavanpodila:
I think a middle ground would be to not throw an AssertionError and instead only print a warning to the console. Can you send a PR, @kholdunn ? If you are not sure, I can help you out.

pavanpodila:
@kholdunn Interesting use case. Can you file an issue for this. I'll work on it next week.

PR #112, opened 2019-02-15
Circle CI

pavanpodila:
This migrates us from TravisCI to CircleCI. The difference in build times is tremendous. Imagine going from **30mins** --> **3mins**!


.circleci/config-local.yml
```diff
@@ -0,0 +1,50 @@
+version: 2
```
pavanpodila: This file is just for trying CircleCI locally using the circle CLI

flutter_mobx/example/pubspec.yaml
```diff
@@ -2,73 +2,5 @@ name: flutter_mobx_example
 description: Examples for flutter_mobx.
```
pavanpodila: All examples have moved to `mobx_examples`

PR #119, opened 2019-02-24
Codegen support for generic stores

### LINKED ISSUE #118: Mobx codegen ignores stores with generic parametes
Example:
```

part 'test_store.g.dart';

class TestStore<T> = TestStoreBase<T> with _$TestStore<T>;

abstract class TestStoreBase<T> implements Store {
  //
}
```
is skipped, nothing is generated, no error in generator output

katis:
Support codegen for generic stores:

```dart
class Item<A> = _Item<A> with _$Item<A>;

abstract class _Item<T> implements Store {
  @observable
  T value;
}
```

Fixes #118

### DIFF
--- a/mobx_codegen/lib/src/template/store.dart
+++ b/mobx_codegen/lib/src/template/store.dart
@@ -1,12 +1,18 @@
 import 'package:mobx_codegen/src/template/action.dart';
 import 'package:mobx_codegen/src/template/async_action.dart';
+import 'package:mobx_codegen/src/template/comma_list.dart';
 import 'package:mobx_codegen/src/template/computed.dart';
 import 'package:mobx_codegen/src/template/observable.dart';
 import 'package:mobx_codegen/src/template/observable_future.dart';
 import 'package:mobx_codegen/src/template/observable_stream.dart';
+import 'package:mobx_codegen/src/template/params.dart';
 import 'package:mobx_codegen/src/template/rows.dart';
 
 class StoreTemplate {
+  final SurroundedCommaList<TypeParamTemplate> typeParams =
+      new SurroundedCommaList('<', '>', []);
+  final SurroundedCommaList<String> typeArgs =
+      new SurroundedCommaList('<', '>', []);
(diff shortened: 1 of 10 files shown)


mobx_codegen/lib/src/template/store.dart
```diff
@@ -1,12 +1,18 @@
...
+import 'package:mobx_codegen/src/template/comma_list.dart';
 import 'package:mobx_codegen/src/template/computed.dart';
 import 'package:mobx_codegen/src/template/observable.dart';
 import 'package:mobx_codegen/src/template/observable_future.dart';
 import 'package:mobx_codegen/src/template/observable_stream.dart';
+import 'package:mobx_codegen/src/template/params.dart';
 import 'package:mobx_codegen/src/template/rows.dart';
 
 class StoreTemplate {
+  final SurroundedCommaList<TypeParamTemplate> typeParams =
```
pavanpodila: Might also need to check for type-parameters for fields, eg: `ObservableList<T>` where the `T` comes from the class type param.

katis: The PR just changes the way codegen detects the class the mixin is applied to, and adds generic params to the mixin definition. Class body is unaffected, and using the generic params in fields etc. Just Works™

alecorsino:
Unfortunately adding JsonSerializable and Generics to the mix it just doesn't work

Here's something i can't make it to work
If I remove Generics it works just fine

`@JsonSerializable()
class Group extends GroupBase with _$Group {
Group({int id, String lines, T type}) : super(id: id, lines: lines, type: type);

factory Group.fromJson(Map<String, dynamic> json) => _$GroupFromJson(json);
Map<String, dynamic> toJson() => _$GroupToJson(this);
}

abstract class GroupBase implements Store {
@observable
int id;

@jsonkey(fromJson: _typeFromJson, toJson: _typeToJson)
@observable
T type;

@observable
String lines;
GroupBase({this.id, this.lines, this.type});
}
`

PR #148, opened 2019-04-12
Update provider example to 2.0

rrousselGit:
With the 2.0-dev version of `provider`, `StatefulProvider` and `Provider` fused.

This PR reflects this change.

### DIFF
--- a/docs/src/guides/organizing-stores.mdx
+++ b/docs/src/guides/organizing-stores.mdx
@@ -134,12 +134,10 @@ Using the [provider][1] package on pub, you can setup a repository of stores at 
 In the example below, we are setting up a store at the app-level:
 
 ```
-final MultiCounterStore store = MultiCounterStore();
-
 class MyApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) => Provider<MultiCounterStore>(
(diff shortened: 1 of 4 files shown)


rrousselGit:
Good point, I forgot about them

rrousselGit:
The existing `pubspec.yaml` doesn't have a version constraint so I assumed it wasn't needed.
But I added it

pavanpodila:
Ah ok. My bad. I think we are ok without the version.

rrousselGit:
Actually, I've just tested and it seems to be required.
It seems that `provider:` resolves to the latest stable release. So it resolves to `1.6.1` instead of `2.0.0-dev`

pavanpodila:
trying to merge, but looks like some glitch in CircleCI, which is stopping it

rrousselGit:
What can I do to unlock the CI?

pavanpodila:
How about pushing some minor change to retrigger?

rrousselGit:
It doesn't seem to work.  🤷‍♂

pavanpodila:
How about closing this PR and creating a new one with a different branch name?

PR #160, opened 2019-04-29
Provider v2

rrousselGit:


docs/src/guides/organizing-stores.mdx
```diff
@@ -178,8 +176,8 @@ class CounterListPage extends StatelessWidget {
 
 There are multiple types of `Providers` available in the [provider][1] package. The ones most relevant and useful to MobX are:
```
pavanpodila: I think we should mention that it should be at least version 2.0 of provider.

docs/src/guides/organizing-stores.mdx
```diff
@@ -134,12 +134,10 @@ Using the [provider][1] package on pub, you can setup a repository of stores at
 In the example below, we are setting up a store at the app-level:
 
 ```
-final MultiCounterStore store = MultiCounterStore();
-
 class MyApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) => Provider<MultiCounterStore>(
-      value: store,
+      builder: (_) => MultiCounterStore(),
```
pavanpodila: Also I am thinking of going for a `MultiProvider`, as each example may have something to "provide" to others deeper in the example widget tree

pavanpodila:
re-opening to see if CircleCI responds

docs/src/guides/organizing-stores.mdx
rrousselGit: Was about to do that but I realized: where is coming from the `examples` variable used on `routes` argument of `MaterialApp`?

Issue #174, opened 2019-05-10
mobx_codegen: allow generating toString method for Store class


atsu85:

pavanpodila:
Curious what would be the implementation of the `toString()` method ?

atsu85:
smth simple like built_value (based on core state fields) would be good enough for me

rrousselGit:
Interesting idea.
Flutter likes to have custom toString to make debugging easier.

For example, instead of `MyClass`, we have `MyClass(foo: 42, bar: 'hello world')`

atsu85:
that format would be ok as well, even tough personally i'd prefer built_value format (for development-time logging), because built_value formats properties on separate lines and excluded nulls :). 
Ideally the format could be customized based on annotation parameter :)

pavanpodila:
If you need a custom toString(), you can always add that to the concrete Store class:

```dart
class CustomStore extends _CustomStore with _$CustomStore {
  @override
  toString() { /* ... */ }
}
```

atsu85:
@pavanpodila,
> If you need a custom toString(), you can always add that to the concrete Store class:

yes, that's the manual approach I've used so far :)

> we can add this to the backlog for generating a toString() implementation with the observables and computeds. Would that work ?

Yes, absolutely. I'd probably prefer just observables (core state) to be included in generated toString, but I see that some might prefer to include computeds as well. Ideally inclusion of computeds could be determined based on annotation parameter.

pavanpodila:
@atsu85, do you want to take a stab at it? All changes are in the `mobx_codegen` package. I can help you tackle it and you will get to add your name to the esteemed list of contributors 🚀

shyndman:
We could go even further, and make the generated class `Diagnosticable` (if the store is marked with a `@forFlutter` annotation or something)

That way we'd get all sorts of debugging goodness in Observatory.

More significant project though.

hawkbee1:
Wanted to try this "good first issue" but..  is this enhancement still needed ?
I see code has changed a lot since May and it looks like dispose method (#177) has been removed from Store (#334) .

pavanpodila:
I think it's still worth having a useful `toString()`, so go all for it 👍. Yes, a few things have changed, but you can certainly add the `toString()` to the mixin.

shyndman:
Maybe as an opt-in via the annotation?

`@Store(generateToString: true)`

pavanpodila:
[...]

PR #177, opened 2019-05-22
Add dispose method to Store

### LINKED ISSUE #176: Should Store interface have a `dispose` method?
This is about the `Store` class from: https://github.com/mobxjs/mobx.dart/blob/5619ed2e21d717072da50febe9615fe2e6aee8e2/mobx/lib/src/api/annotations.dart#L25

This class is used as a base class for all custom mobx stores. As such, it is logical for libraries built on the top of `Mobx` to use the `Store` interface for object constraints.

Example:

```dart
class Foo<T extends Store> {
  // TODO: do something Mobx specific with T
}
```

At the same time, while many stores don't need to be disposed of, many do due to internal reactions like so:

```dart
class MyStore implements Store {
  ReactionDisposer _disposer;

  MyStore() {
    _disposer = when(something);
  }

  void dispose() {
    _disposer();
  }
}
```

____

The problem is, if library makers want to interact with stores (for example controlling when they are created/disposed), then these libraries won't be able to correctly dispose of stores with internal reactions.

So if `Store` looked more like this:

```dart
abstract class Store {
  void dispose() {}
}
```

That would allow libraries to make something like this:

```dart
class Example<T extends Store> extends StatefulWidget {
  const Example({Key key, this.builder}) : super(key: key);

  final T Function() builder;

  @override
  _ExampleState<T> createState() => _ExampleState<T>();
}

class _ExampleState<T extends Store> extends State<Example<T>> {
  T store;

  @override
  void initState() {
    super.initState();
    store = widget.builder();
  }

  @override
  void dispose() {
    store.dispose();
    super.dispose();
  }
}
```

Where this class handles both creation and cleaning up of a `Store`.

This is something that I'd like as an attempt to solve https://github.com/mobxjs/mobx.dart/issues/161 with `provider`.

rrousselGit:
closes #176

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ part 'counter.g.dart';
 
 class Counter = CounterBase with _$Counter;
 
-abstract class CounterBase implements Store {
+abstract class CounterBase with Store {
   @observable
   int value = 0;
 
@@ -140,7 +140,7 @@ part 'contact.g.dart';
 
 class Contact = ContactBase with _$Contact;
 
-abstract class ContactBase implements Store {
+abstract class ContactBase with Store {
   @observable
   String firstName;
 
@@ -183,7 +183,7 @@ part 'counter.g.dart';
 
 class Counter = CounterBase with _$Counter;
 
-abstract class CounterBase implements Store {
+abstract class CounterBase with Store {
   @observable
   int value = 0;
 
@@ -299,7 +299,7 @@ part 'counter.g.dart';
 
 class Counter = CounterBase with _$Counter;
 
-abstract class CounterBase implements Store {
+abstract class CounterBase with Store {
   @observable
   int value = 0;
 
--- a/docs/src/api/action.mdx
+++ b/docs/src/api/action.mdx
@@ -44,7 +44,7 @@ part 'counter.g.dart';
 
 class Counter = _Counter with _$Counter;
 
-abstract class _Counter implements Store {
+abstract class _Counter with Store {
   @observable
   int value = 0;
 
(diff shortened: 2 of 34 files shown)


rrousselGit:
Note that this is breaking change even without the `implements` -> `with` change.

rrousselGit:
Another point is, we could allow both:

- `class MyStore with Store`
- `class MyStore implements Store { void dispose() {} }`

With the current changes, only the former is possible.

We could allow both, but I think it is interesting to forces using `with`. This way in the future if other methods are added on `Store`, this won't be a breaking change.

rrousselGit:
What's the purpose of `Store` interface then?

pavanpodila:
I was thinking of isolating the "dispose" behavior into an interface. But looks like this is an intrinsic behavior and should really be part of the interface/mixin. Someone else was also asking about a `toString` implementation (#174). I guess it would be easier to do it then.

Since this is a breaking change, we should change our version as well accordingly.

rrousselGit:
I don't understand how this is related to `toString`.

Not do I understand what I have to do now haha

pavanpodila:
LoL. My question was about versioning. Since it's a breaking change we will need some docs on the change. I'll try it out today and then merge. Busy with some client work.

talamaska:
Hi, I'm not sure what happened with this change, but it seems in 2021 this is gone from the Store mixin.
Should I just implement a normal action called dispose and handle whatever I need to dispose?

anton-brass:
what is the answer of @talamaska  question?

PR #187, opened 2019-06-05
Update README.md - async actions

yoavrofe:


README.md
```diff
@@ -194,6 +194,26 @@ abstract class CounterBase with Store {
 }
 ```
 
+#### Asynchronous Actions
+
+MobX.dart handles asynchronous actions [automatically](https://github.com/mobxjs/mobx.dart/pull/89) and does not require wrapping the code with ```runInAction```.
```
pavanpodila: I think we should skip the part about mentioning the pull request. Too much detail for our users ;-). Also the `runInAction` should link to that section on API. There is no need to use the triple quoted string. Instead a plain backquoted string should be good enough, such as \`runInAction\`

yoavrofe: Thanks. I updated it but for some reason cannot see the changes in the PR.

yoavrofe: I see them now.

PR #194, opened 2019-06-10
fix observable list notifications and improve test coverage

### LINKED ISSUE #193: When adding an item to an ObservableList and it is observed, the index is length of the list
I've defined a store as follows

```
import '../models/item.dart';

part 'data_store.g.dart';

class DataStore = _DataStore with _$DataStore;

abstract class _DataStore with Store {
  @observable
  ObservableList<Item> items = ObservableList<Item>();

  @action
  void addItem() {
    items.add(Item(DateTime.now().toUtc()));
  }
}
```

In one of my widgets, I observe if the list has been modified

```
@override
  void initState() {
    super.initState();
    widget.store.items.observe((listChange) {
      if (listChange.added?.isNotEmpty ?? false) {
        <handle when item has been added>
      }
    });
  }
```

When an item has been added via the `addItem` method within the store, the callback is invoked in the widget to notify me that an item has been added. Upon inspect the value of `listChange.index`, it appears that rather than being the index of where the item has been added, it's actually the length of the list. For example, if this is the first item that's been added, I'm expecting `listChange.index` to have a value of `0`, instead the actual value is `1`

Provided it's an actual bug (which I suspect is the case), [this](https://github.com/mobxjs/mobx.dart/blob/f337f10b03a103f47786471fae6b041ce4d45a51/mobx/lib/src/api/observable_collections/observable_list.dart#L93) appears to be where the fix needs to be applied

Edit: code snippets are taken from this repo  https://github.com/MaikuB/flutter_mobx_animatedlist_demo if you need to be able to reproduce the issue. I can look at submitting a PR as well though should be an easy fix

MaikuB:
This should fix #193. As part of this, it looked like the `addAll` and `replaceRange` had similar issues so made some changes there. Added in some extra tests to improve the test coverage as well

### DIFF
--- a/mobx/lib/src/api/observable_collections/observable_list.dart
+++ b/mobx/lib/src/api/observable_collections/observable_list.dart
@@ -88,17 +88,17 @@ class ObservableList<T>
   @override
   void add(T element) {
     _context.checkIfStateModificationsAreAllowed(_atom);
-
+    final index = _list.length;
     _list.add(element);
-    _notifyListUpdate(_list.length, [element], null);
+    _notifyListUpdate(index, [element], null);
   }
 
   @override
   void addAll(Iterable<T> iterable) {
     _context.checkIfStateModificationsAreAllowed(_atom);
-
+    final index = _list.length;
     _list.addAll(iterable);
-    _notifyListUpdate(0, iterable.toList(growable: false), null);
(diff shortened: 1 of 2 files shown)


pavanpodila:
Thanks a lot @MaikuB 👍

MaikuB:
No worries :)

PR #195, opened 2019-06-10
consolidate .gitignore

MaikuB:
not sure if intentional or not but it looks like the folder for the mobx package doesn't have a .gitignore file. took the one from mobx_codegen that should work for the mobx package as well. feel free to close this if leaving out the .gitignore was intentional


pavanpodila:
We have one `.gitignore` at the top most level, hence there is no need for per-project ones. I think I should remove the per-project `.gitignore` as well. May you be can help in removing it :-)

MaikuB:
Ah right. Thanks for pointing that out. I've taken a stab at this. Hopefully I haven't missed something. One thing I noticed is it looks like mobx_examples is suppose to be a standalone Flutter app with examples but is missing the iOS and Android project file. Unless I'm missing something, anyone that clones the repo and goes to that folder won't be able to run the examples as is

pavanpodila:
Ya, we had a discussion about the `ios` and `android` folders under `mobx_examples` sometime back. The suggestion was to not keep it as its mostly generated code with no customization such as icons/language-dictionary/project-settings.

These folders can be generated by simply running
```shell
flutter create .
```
inside the `mobx_examples` folder. Since then, I've added these folders to the `.gitignore`. If you think that we should keep it, I would love to know the rationale.

MaikuB:
Ah yep that forgot about that command. My thinking was just that it helps users being able to run the examples sooner. Unless I've missed it, there's no mention to run the command in the docs. Most packages/example apps include those already so it's probably not unreasonable for a member of the community to default to expecting that the files are there so they could run the app.

I'm also wondering if it's possible for a scenario to occur where when the users run the command and then run into issues when they run it. For example, it ends up compiling against a different version of the Android SDK that has issues with bringing up the keyboard in the form example. It's very unlikely these scenarios can happen, though there was an issue with keyboards in the past

PR #205, opened 2019-06-21
Enforce reads

pavanpodila:
Introduces a new setting on the `ReactiveConfig`: `readPolicy`. With this setting it checks if the reactive-reads are happening outside of an action / reaction. When `ReactiveReadPolicy.always` is used, it will throw an exception for stray reads of the observable. This prevents some logic errors when a reactive read happens outside of a reactive context like an action / reaction.

There are also some renames of earlier enums.

`EnforceActions` => `ReactiveWritePolicy`


rrousselGit:
Should le enforce read logic performed only in release mode, through asserts?

pavanpodila:
> Should le enforce read logic performed only in release mode, through asserts?

You mean `assert()` in debug mode ? That's a good point. I think if we use asserts, then the `writePolicy` should also be treated as an `assert()` ?

rrousselGit:
That make sense for `writePolicy` too, indeed

rrousselGit:
If we do want to use asserts to remove the related piece of code in release, then it may be a good idea to rename the variables to `debugReadPolicy`.

That's the kind of name Flutter/Dart use to express that a variable is for devopment only.

pavanpodila:
> If we do want to use asserts to remove the related piece of code in release, then it  I tmay be a good idea to rename the variables to `debugReadPolicy`.
> 
> That's the kind of name Flutter/Dart use to express that a variable is for devopment only.

Little hesitant over here since the read/write-policy could be enforced at runtime too. Possibly in a future release. MobX.js does this at runtime and this is leveraged by MobX State Tree to enforce  typing. I think we should wait it out and see how the community uses it ? For now, I'll keep it as asserts and also keep the name without the `debug`-prefix.

Thoughts?

rrousselGit:
Sure, I don't mind.
But having the ability to perform a runtime check also means that dart won't be able to remove the related code of the feature is disabled.

pavanpodila:
We will keep it as asserts for now so there is no runtime cost. Only the naming of the fields stay.

pavanpodila:
@rrousselGit Can you take a look now? I also need to update docs

mobx/lib/src/api/observable_collections/observable_list.dart
```diff
@@ -48,34 +48,40 @@ class ObservableList<T>
 
   @override
   int get length {
+    _context.enforceReadPolicy(_atom);
```
rrousselGit: All calls to `enforeReadPolicy` should also be wrapped in an assert like so:

```dart
assert(() {
  _context.enforeReadPolicy(_atom);
  return true;
}());
```

Otherwise, it won't be properly tree-shaked, especially since `enforeReadPolicy` does more than just an assert.

That's one of the reasons they use `debug` in the name, to make it easier to see that all calls should be wrapped in such assert.

pavanpodila: Sure, I can do that. Only thing in case of `enforceWritePolicy`, it can't be wrapped as there is a consistency check as well, which is not an assert. If we keep this wrapping inside `enforceReadPolicy`, then for release the method becomes empty with no statement. Will it be tree-shaked to remove the empty method?

We will wrap inside with an `assert`.
```dart
void enforceWritePolicy(Atom atom) {
  assert((){ /* ... */ }());
}
```

During release build, it should become empty:

```dart
void enforceWritePolicy(Atom atom) {
}
```

Will this be tree-shaked as well to remove such empty methods ? If that's the its easier to centralize then to sprinkle the assert wrapper everywhere.

rrousselGit: I'm not sure about that. It's not very easy to test.
But I know for a fact that flutter wraps both the implementation and the call in an `assert`.

pavanpodila: Ok, I can apply that wrapper for both read/write. Just for consistency I'll keep it inside the `enforce***` methods. I know it will still result in empty methods but we will not be executing the code at runtime. Hope that's a decent compromise without sacrificing too much readability.

rrousselGit: That's fine.

pavanpodila: I've applied the assert wrapped but kept it inside the `enforce***` methods.

pavanpodila:
@rrousselGit Need your help in diagnosing why 3 of tests in `mobx_codegen` are failing. Debugging through this I saw that `testBuilder` (part of `build_test` package) is not generating any output and the `StoreGenerator` is never getting hit. Can you check and advise how to resolve this?

rrousselGit:
I'll take a look

rrousselGit:
From my tests, `StoreGenerator` does get hit. But for some reasons, the generator runs on `package:mobx/src/api/annotations.dart` instead of the input sources.

rrousselGit:
I'm not exactly sure what the cause of the issue is.

I'll continue looking into it, but just in case:

A few months ago the Dart team extracted its testing tool used for `json_serializable` into a package: https://github.com/kevmoo/source_gen_test. I use a fork on `functional_widget` too. 

It takes care of all the things that the mocked sources and `generate` do.
That's a potential solution to the problem.

pavanpodila:
For me the StoreGenerator is not even getting hit. Not sure if there is some environment issue here. I am running on flutter stable right now.

BTW, do you use the dart-sdk from flutter or have you installed Dart separately?

pavanpodila:
> I'm not exactly sure what the cause of the issue is.
> 
> I'll continue looking into it, but just in case:
> 
> A few months ago the Dart team extracted its testing tool used for `json_serializable` into a package: https://github.com/kevmoo/source_gen_test. I use a fork on `functional_widget` too.
> 
> It takes care of all the things that the mocked sources and `generate` do.
> That's a potential solution to the problem.

Ah nice, let me explore this a bit

rrousselGit:
I use the VM Dart SDK. 

I had to change the version of `build_test` though. The `reader` argument is available only in the 0.10.8, but the current version used is 0.10.7

pavanpodila:
> I use the VM Dart SDK.
> 
> I had to change the version of `build_test` though. The `reader` argument is available only in the 0.10.8, but the current version used is 0.10.7

Ok..I am sticking to the flutter dart-sdk as that's the one used in my teams. Hope its not too different, although I am might be some versions behind

rrousselGit:
> I am sticking to the flutter dart-sdk

You can have both coexisting.
I'm not sure if using Flutter's Dart sdk will properly work. Code-generation use dart:mirror at some points.

pavanpodila:
> > I am sticking to the flutter dart-sdk
> 
> You can have both coexisting.
> I'm not sure if using Flutter's Dart sdk will properly work. Code-generation use dart:mirror at some points.

Interestingly it was working until this PR :-). Not sure what changed besides a few renames of the methods.

rrousselGit:
I think I found the issue.

It's the line:

```dart
c.mixins.any(_storeChecker.isExactlyType);
```
that always returns false.

It seems like the checker doesn't match with the mocked sources of `Store`.

Using the following instead fixes the tests

```dart
c.mixins.any((c) => c.displayName == "Store");
```

Although that's a workaround.

I would expect the package mentioned earlier for code-gen tests to fix that issue.

pavanpodila:
I think I got this fixed now. The problem was with the fake sources. After removing them and using the `PackageAssetReader`, the Store type is correctly resolved and passes all checks. Your last comment gave me the real hint!

PR #241, opened 2019-07-30
Reaction comparator

peteyycz:
I'm already using this feature in my project but I'd like to get some feedback on this, if I'm doing it right I'd also like to contribute more with docs & tests. 😄

Add the possibility to override equality comparator function in reactions. This feature is already present in the JavaScript version, detailed here: https://mobx.js.org/refguide/reaction.html

The problem originally was that I have 2 types which I don't have control of, to override the `compareTo` or `==` operator, but they are constantly changing in my app and I need to react on changes of them even if they are just reassigned with the same value.

What do you think, is this feature a step towards the right direction or should mobx encourage people defining their own wrappers and equality?

EDIT: Update after some investigation I realized that this cannot be added for `autorun` THAT easily 😄  and does not really make sense for `when` nor `asyncWhen`.

### DIFF
--- a/mobx/test/observe_test.dart
+++ b/mobx/test/observe_test.dart
@@ -36,6 +36,60 @@ void main() {
       dispose();
     });
 
+    group('equality override', () {
+      test('yields a new value', () {
+        final x = Observable(
+          10,
+          equals: (_, __) => false,
+        );
+
+        var executed = false;
+
+        final dispose = x.observe((change) {
+          expect(change.newValue, equals(10));
+          executed = true;
+        }, fireImmediately: true);
+
+        expect(executed, isTrue);
+        executed = false;
+
+        x.value = 10;
+        expect(executed, isTrue);
+        executed = false;
+
+        x.value = 10;
+        expect(executed, isTrue);
+        executed = false;
+
+        dispose();
+      });
+
+      test('does not yield a new value', () {
+        final x = Observable(
+          10,
+          equals: (_, __) => true,
+        );
+
+        var executed = false;
+
+        final dispose = x.observe((change) {
+          expect(change.newValue, equals(10));
+          executed = true;
+        }, fireImmediately: true);
+
+        expect(executed, isTrue);
+        executed = false;
+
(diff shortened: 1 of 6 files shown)


pavanpodila:
Hey @peteyycz, Sorry about the late response. Was too caught up at work. I think this is a good feature to have especially for expensive reactions. Also don't think `autorun` or `when` needs it. I am ok with just having it on a `reaction`

mobx/test/reaction_test.dart
```diff
@@ -35,6 +35,40 @@ void main() {
...
+            executed = true;
+          },
+          name: 'Basic Reaction',
+          equals: equals,
+        );
+
+        expect(executed, isFalse);
+
+        x.value = 11;
+        expect(executed, isTrue);
```
pavanpodila: Not sure how this is working as the comparator is always returning false. Hence the effect function should never execute?

peteyycz: It's the other way around, the original implementation had a case for _not_ running when the items are _equal_, that was an optimization to not run extra functions for the same values.

mobx/test/observe_test.dart
```diff
@@ -36,6 +36,60 @@ void main() {
...
+
+        final dispose = x.observe((change) {
+          expect(change.newValue, equals(10));
+          executed = true;
+        }, fireImmediately: true);
+
+        expect(executed, isTrue);
+        executed = false;
+
+        x.value = 10;
```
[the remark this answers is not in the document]
peteyycz: See my comment below.

mobx/test/reaction_test.dart
pavanpodila: Ah ok. Got it. The snippet out of context confused me a bit 🤔.

mobx/test/observe_test.dart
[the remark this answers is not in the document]
pavanpodila: I think it would be good to put in a few comments around why we are assigning the same value and not changing it. Good for a future self 😉

PR #243, opened 2019-08-13
Update community page

MaikuB:
I've changed domains (though the old one should redirect). Paul's article no longer exists on his site so added a link to the video. Also removed the periods that were being used inconsistently


pavanpodila:
Thanks Michael 👍

Issue #225, opened 2019-07-12
Simplify Store creation with the @store annotation


pavanpodila:
Based on the discussion in #217, we can simplify the Store creation with the use of an annotation: `@store`.

The ideal way given the code-gen constraints is to have something like this:

```dart
part 'counter.g.dart'; // This file will contain the _$Counter mixin

@store
class Counter with _$Counter {
  @observable
  int value = 0;

  @action
  void increment() {
    value++;
  }

}
```

The generated code will be as before with some slight changes on the mixin declaration:

```dart
// GENERATED CODE - DO NOT MODIFY BY HAND

part of 'counter.dart';

// **************************************************************************
// StoreGenerator
// **************************************************************************

// ignore_for_file: non_constant_identifier_names, unnecessary_lambdas, prefer_expression_function_bodies, lines_longer_than_80_chars

mixin _$Counter on Counter {
  final _$valueAtom = Atom(name: '_Counter.value');

  @override
  int get value {
    _$valueAtom.context.enforceReadPolicy(_$valueAtom);
    _$valueAtom.reportObserved();
    return super.value;
  }

  @override
  set value(int value) {
    // Since we are conditionally wrapping within an Action, there is no need to enforceWritePolicy
    _$valueAtom.context.conditionallyRunInAction(() {
      super.value = value;
      _$valueAtom.reportChanged();
    }, _$valueAtom, name: '${_$valueAtom.name}_set');
  }

  final _$_CounterActionController = ActionController(name: '_Counter');

  @override
  void increment() {
    final _$actionInfo = _$_CounterActionController.startAction();
    try {
      return super.increment();
    } finally {
      _$_CounterActionController.endAction(_$actionInfo);
    }
  }
}

```

@rrousselGit, can you confirm this is doable?

rrousselGit:
It's not.
We need that `class Foo = Base with $Generated`. 

Mixins naturally don't override members of a class, which mobx needs to do for custom getter/setter

pavanpodila:
Ah. So we are not really saving much, but the `@store` annotation is a good start.

rrousselGit:
That's a lot more reasonable yes.

If we're going for tmsuch change, would it be possible to force `_$Base` to be `_$BaseStore`?

That would allow:

```
class Foo = Base with _$BaseStore, _$AnotherGeneratedMixin
```

pavanpodila:
That might be a problem if you are defining multiple stores in the same file.

[...]

PR #245, opened 2019-08-15
Simplify class matching logic in codegen

shyndman:
While prepping support for adding `@store` annotation support, I saw an
opportunity to generalize how we match on suitable base class / mixed
class pairs.

The insight was that the two paths through the match logic (type 
parameterized vs not) were essentially the same, if you consider
a unparameterized type the same as an parameterized type with zero type 
params.

I also added a bunch of comments, which should help the next person
digest it.

Related to #225

### DIFF
--- a/mobx_codegen/lib/src/mobx_codegen_base.dart
+++ b/mobx_codegen/lib/src/mobx_codegen_base.dart
@@ -24,34 +24,41 @@ class StoreGenerator extends Generator {
 
   @override
   FutureOr<String> generate(LibraryReader library, BuildStep buildStep) {
-    Iterable<String> generate(ClassElement baseClass) sync* {
-      final mixedClass =
-          library.classes.where((c) => c != baseClass).firstWhere((c) {
-        if (c.supertype.typeArguments.isNotEmpty &&
-            baseClass.typeParameters.length ==
-                c.supertype.typeArguments.length) {
-          final t = baseClass.type.instantiate(c.supertype.typeArguments);
-          return t.isSupertypeOf(c.type);
-        }
-        return c.supertype == baseClass.type;
-      }, orElse: () => null);
-
-      if (mixedClass != null) {
-        yield generateStoreClassCode(library, baseClass, mixedClass);
-      }
-    }
-
     return library.classes
         .where(isValidStoreClass)
-        .expand(generate)
+        .expand((c) => expandToMixinCode(library, c))
         .toSet()
         .join('\n\n');
+  }
+
+  Iterable<String> expandToMixinCode(
+    LibraryReader library,
+    ClassElement baseClass,
+  ) sync* {
+    final otherClasses = library.classes.where((c) => c != baseClass);
+    final mixedClass = otherClasses.firstWhere((c) {
+      // If our base class has different type parameterization requirements than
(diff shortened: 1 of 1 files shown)


mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -24,34 +24,43 @@ class StoreGenerator extends Generator {
...
+    final mixedClass = otherClasses.firstWhere((c) {
+      // If our base class has different type parameterization requirements than
+      // the class we're evaluating provides, we know it's not a subclass.
+      if (baseClass.typeParameters.length != c.supertype.typeArguments.length) {
+        return false;
+      }
+
+      // Apply the subclass' type arguments to the base type, and perform a
+      // supertype check.
+      return baseClass.type
```
shyndman: Just had a thought — this might throw if the type parameters impose constraints.

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -24,34 +24,43 @@ class StoreGenerator extends Generator {
...
+          .instantiate(c.supertype.typeArguments)
```
shyndman: Just had a thought — this might(?) throw if the type parameters impose constraints.

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -24,34 +24,43 @@ class StoreGenerator extends Generator {
...
+      // the class we're evaluating provides, we know it's not a subclass.
+      if (baseClass.typeParameters.length != c.supertype.typeArguments.length) {
+        return false;
+      }
+
+      // Apply the subclass' type arguments to the base type (if there are none
+      // this has no impact), and perform a supertype check.
+      return baseClass.type
+          .instantiate(c.supertype.typeArguments)
+          .isDirectSupertypeOf(c.type);
```
shyndman: `isDirectSupertypeOf` is probably a better pick than `isSupertypeOf`, which was being used before. It's a closer match to the equality check that was being performed in the absence of type params.

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -24,34 +24,41 @@ class StoreGenerator extends Generator {
...
```
pavanpodila: `isDirectSupertypeOf` is now deprecated: https://pub.dev/documentation/analyzer/latest/dart_element_type/InterfaceType/isDirectSupertypeOf.html. Better to go back to `isSupertypeOf`

shyndman: Done

shyndman:
PTAL.

Btw, What's up with the checks?

shyndman:
I'm closing this off until the `@store` annotation is merged.

I'm used to working in a chained CL workflow, and GH doesn't seem to support it very well.

shyndman:
@pavanpodila If you have any concerns about this, let me know. Otherwise, let's get this merged.

pavanpodila:
Are the tests passing for mobx_codegen ? :-)

shyndman:
yep

PR #247, opened 2019-08-16
Store annotation codegen

shyndman:
Adds code generation support for `@store` annotation syntax.

If a private class annotated with `@store` is encountered, we now generate its public, MobX counterpart.

Example:
```dart
@store
class _User {
  @observable
  int value;
}
```

Produces:
```dart
class User extends _User {
  final _$valueAtom = Atom(name: '_User.value');

  @override
  int get value {
    _$valueAtom.context.enforceReadPolicy(_$valueAtom);
    _$valueAtom.reportObserved();
    return super.value;
  }

  @override
  set value(int value) {
    _$valueAtom.context.conditionallyRunInAction(() {
      super.value = value;
      _$valueAtom.reportChanged();
    }, _$valueAtom, name: '${_$valueAtom.name}_set');
  }
}'
```

We do not remove support for the existing, mixin-style syntax.

### DIFF
--- a/mobx_codegen/lib/src/errors.dart
+++ b/mobx_codegen/lib/src/errors.dart
@@ -6,6 +6,7 @@ abstract class CodegenError {
 class StoreClassCodegenErrors implements CodegenError {
   StoreClassCodegenErrors(this.name) {
     _errorCategories = [
+      invalidStoreDeclaration,
       invalidComputedAnnotations,
       invalidObservableAnnotations,
       invalidActionAnnotations,
@@ -19,6 +20,8 @@ class StoreClassCodegenErrors implements CodegenError {
 
   final String name;
 
+  final InvalidStoreDeclarations invalidStoreDeclaration =
+      InvalidStoreDeclarations();
   final PropertyErrors finalObservables = FinalObservableFields();
   final PropertyErrors staticObservables = StaticObservableFields();
   final PropertyErrors staticMethods = InvalidStaticMethods();
@@ -52,6 +55,26 @@ class StoreClassCodegenErrors implements CodegenError {
 const fieldPluralizer = Pluralize('the field', 'fields');
 const methodPluralizer = Pluralize('the method', 'methods');
 const memberPluralizer = Pluralize('the member', 'members');
+
+class InvalidStoreDeclarations implements CodegenError {
+  final NameList _classNames = NameList();
+
+  // ignore: avoid_positional_boolean_parameters
+  bool addIf(bool condition, String className) {
+    if (condition) {
+      _classNames.add(className);
+    }
+    return condition;
+  }
+
+  @override
+  bool get hasErrors => _classNames.isNotEmpty;
+
+  @override
+  String get message => 'Store classes cannot be defined with both the @store '
+      'annotation and Store mixin. Please choose one method or another. '
+      '$_classNames';
+}
 
 abstract class PropertyErrors implements CodegenError {
   final NameList _properties = NameList();
--- a/mobx_codegen/lib/src/mobx_codegen_base.dart
+++ b/mobx_codegen/lib/src/mobx_codegen_base.dart
@@ -6,7 +6,7 @@ import 'package:build/build.dart';
 import 'package:mobx/mobx.dart' show Store;
 // ignore: implementation_imports
 import 'package:mobx/src/api/annotations.dart'
-    show ComputedMethod, MakeAction, MakeObservable;
+    show ComputedMethod, MakeAction, MakeObservable, MakeStore;
 import 'package:mobx_codegen/src/errors.dart';
 import 'package:mobx_codegen/src/template/action.dart';
 import 'package:mobx_codegen/src/template/async_action.dart';
@@ -20,17 +20,21 @@ import 'package:mobx_codegen/src/template/util.dart';
 import 'package:source_gen/source_gen.dart';
 
 class StoreGenerator extends Generator {
-  final _storeChecker = const TypeChecker.fromRuntime(Store);
-
   @override
   FutureOr<String> generate(LibraryReader library, BuildStep buildStep) =>
-      library.classes
-          .where(isValidStoreClass)
-          .expand((c) => expandToMixinCode(library, c))
-          .toSet()
-          .join('\n\n');
+      _generateCodeForLibrary(library).toSet().join('\n\n');
 
-  Iterable<String> expandToMixinCode(
+  Iterable<String> _generateCodeForLibrary(LibraryReader library) sync* {
+    for (final classElement in library.classes) {
+      if (isValidMixinStoreClass(classElement)) {
+        yield* _generateCodeForMixinStore(library, classElement);
+      } else if (isValidAnnotatedStoreClass(classElement)) {
+        yield* _generateCodeForAnnotatedStore(library, classElement);
+      }
+    }
+  }
+
+  Iterable<String> _generateCodeForMixinStore(
     LibraryReader library,
     ClassElement baseClass,
   ) sync* {
@@ -50,31 +54,59 @@ class StoreGenerator extends Generator {
     }, orElse: () => null);
 
     if (mixedClass != null) {
-      yield generateStoreMixinCode(library, baseClass, mixedClass);
+      yield _generateCodeFromTemplate(
+          mixedClass.name, baseClass, MixinStoreTemplate());
     }
   }
 
-  bool isValidStoreClass(ClassElement c) =>
-      c.isAbstract && c.mixins.any(_storeChecker.isExactlyType);
+  Iterable<String> _generateCodeForAnnotatedStore(
+    LibraryReader reader,
+    ClassElement baseClass,
+  ) sync* {
+    assert(baseClass.isPrivate);
+    // Strip off leading underscore
+    final publicTypeName = baseClass.name.substring(1);
+    yield _generateCodeFromTemplate(
+        publicTypeName, baseClass, SubclassStoreTemplate());
+  }
 
-  String generateStoreMixinCode(
-      LibraryReader library, ClassElement baseClass, ClassElement mixedClass) {
-    final visitor = StoreMixinVisitor(baseClass, mixedClass.name);
-    baseClass.visitChildren(visitor);
+  String _generateCodeFromTemplate(
+    String publicTypeName,
+    ClassElement userStoreClass,
+    StoreTemplate template,
+  ) {
+    final visitor = StoreMixinVisitor(publicTypeName, userStoreClass, template);
+    userStoreClass
+      ..accept(visitor)
+      ..visitChildren(visitor);
     return visitor.source;
   }
 }
 
+const _storeMixinChecker = TypeChecker.fromRuntime(Store);
+const _storeAnnotationChecker = TypeChecker.fromRuntime(MakeStore);
+
+bool isValidMixinStoreClass(ClassElement classElement) =>
+    classElement.isAbstract &&
+    classElement.mixins.any(_storeMixinChecker.isExactlyType);
+
+bool isValidAnnotatedStoreClass(ClassElement classElement) =>
+    classElement.isPrivate &&
+    _storeAnnotationChecker.hasAnnotationOfExact(classElement);
+
 class StoreMixinVisitor extends SimpleElementVisitor {
-  StoreMixinVisitor(ClassElement parentClass, String name)
-      : _errors = StoreClassCodegenErrors(name) {
-    _storeTemplate = StoreTemplate()
+  StoreMixinVisitor(
+    String publicTypeName,
+    ClassElement userClass,
+    StoreTemplate template,
+  ) : _errors = StoreClassCodegenErrors(publicTypeName) {
+    _storeTemplate = template
       ..typeParams
           .templates
-          .addAll(parentClass.typeParameters.map(typeParamTemplate))
-      ..typeArgs.templates.addAll(parentClass.typeParameters.map((t) => t.name))
-      ..parentName = parentClass.name
-      ..mixinName = '_\$$name';
+          .addAll(userClass.typeParameters.map(typeParamTemplate))
+      ..typeArgs.templates.addAll(userClass.typeParameters.map((t) => t.name))
+      ..parentTypeName = userClass.name
+      ..publicTypeName = publicTypeName;
   }
 
   final _observableChecker = const TypeChecker.fromRuntime(MakeObservable);
@@ -95,6 +127,14 @@ class StoreMixinVisitor extends SimpleElementVisitor {
       return '';
     }
     return _storeTemplate.toString();
+  }
+
+  @override
+  void visitClassElement(ClassElement element) {
+    if (isValidAnnotatedStoreClass(element) &&
+        isValidMixinStoreClass(element)) {
+      _errors.invalidStoreDeclaration.addIf(true, element.name);
+    }
   }
 
   @override
--- a/mobx_codegen/lib/src/template/store.dart
+++ b/mobx_codegen/lib/src/template/store.dart
@@ -8,13 +8,39 @@ import 'package:mobx_codegen/src/template/observable_stream.dart';
 import 'package:mobx_codegen/src/template/params.dart';
 import 'package:mobx_codegen/src/template/rows.dart';
 
-class StoreTemplate {
+class SubclassStoreTemplate extends StoreTemplate {
+  String get typeName => publicTypeName;
+
+  @override
+  String toString() => '''
+  ${StoreTemplate.analyzerIgnores}
+
+  class $typeName$typeParams extends $parentTypeName$typeArgs {
+    $storeBody
+  }''';
+}
+
+class MixinStoreTemplate extends StoreTemplate {
+  String get typeName => '_\$$publicTypeName';
+
+  @override
+  String toString() => '''
+  ${StoreTemplate.analyzerIgnores}
+
+  mixin $typeName$typeParams on $parentTypeName$typeArgs, Store {
+    $storeBody
+  }''';
+}
+
+abstract class StoreTemplate {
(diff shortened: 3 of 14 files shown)


mobx_codegen/pubspec.yaml
```diff
@@ -14,7 +14,8 @@ environment:
 dependencies:
   path: ^1.6.2
   source_gen: ^0.9.4
-  mobx: ^0.3.3
+  mobx:
+    path: ../mobx
```
pavanpodila: This should point to the pub-dev package (`^x.y.z`) as before and not local. Else publishing would not work. You can use the [dependency_overrides](https://dart.dev/tools/pub/dependencies#dependency-overrides) option instead.

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -24,34 +24,41 @@ class StoreGenerator extends Generator {
...
+      // the class we're evaluating provides, we know it's not a subclass.
+      if (baseClass.typeParameters.length != c.supertype.typeArguments.length) {
+        return false;
+      }
+
+      // Apply the subclass' type arguments to the base type (if there are none
+      // this has no impact), and perform a supertype check.
+      return baseClass.type
+          .instantiate(c.supertype.typeArguments)
+          .isDirectSupertypeOf(c.type);
```
pavanpodila: Could you please tell me the pros/cons of `isDirectSupertypeOf` vs `isSupertypeOf`?

mobx/lib/src/api/annotations.dart
```diff
@@ -1,3 +1,10 @@
+class MakeStore {
+  const MakeStore._();
+}
+
+/// Declares a class as a store.
+const MakeStore store = MakeStore._();
```
pavanpodila: You may want to publish an alpha version package of `mobx`, which the `mobx_codegen` can depend on. The alpha version are ignored by pub as they are not included in the version check

pavanpodila: We also need a bunch of tests / examples

shyndman:
Oh hey, @pavanpodila. This PR is draft and marked WIP. I'm going to keep updating until it's done, and then it'll be ready for review. Sound good?

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -24,34 +24,41 @@ class StoreGenerator extends Generator {
...
+    }, orElse: () => null);
+
+    if (mixedClass != null) {
+      yield generateStoreMixinCode(library, baseClass, mixedClass);
+    }
+  }
+
   bool isValidStoreClass(ClassElement c) =>
```
pavanpodila: Will have to revisit this check. We don't have mixins anymore with the `@store` annotation?

pavanpodila:
> Oh hey, @pavanpodila. This PR is draft and marked WIP. I'm going to keep updating until it's done, and then it'll be ready for review. Sound good?

Ok ya. I got a bit excited to see this :-). Will wait for you to complete.

shyndman:
You're welcome to take a look at that other PR. This one is dependent on that.

rrousselGit:
Will we have both versions working together?
And how would we generate a private store?

If we're going for this, it might also be interesting to try to make an analyzer-plugin and/or contact the analyzer team about it.
Issues about renaming and "go to definition" could be solved this way.

shyndman:
Hi @rrousselGit.

I'm writing this so both approaches will be supported. You folks can make the call on whether to deprecate one approach versus the other.

My feeling is that we should get the public subclass working, and see if there's any demand for private. We can definitely figure something out if we need to.

Your call on contacting the analyzer folks. The ergonomics of jumping to the def of a generated class wouldn't bother me, personally. Renaming would be slightly annoying though.

mobx/lib/src/api/annotations.dart
shyndman: Sounds like a plan. I'm going to split this into two PRs. The first to update mobx. Once we have that in, would you mind reporting back once you've published? I'll update this PR's pubspec to match.

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -20,55 +20,87 @@ import 'package:mobx_codegen/src/template/util.dart';
...
-        }
-        return c.supertype == baseClass.type;
-      }, orElse: () => null);
-
-      if (mixedClass != null) {
-        yield generateStoreClassCode(library, baseClass, mixedClass);
+    final generatedMixins = library.classes
+        .where(isValidMixinStoreClass)
+        .expand((c) => generateCodeForMixinStore(library, c));
+    final generatedSubclasses = library.classes
```
pavanpodila: We need to handle the case where both the `@store` and `Store` mixin are used together

mobx_codegen/lib/src/template/store.dart
```diff
@@ -8,35 +8,38 @@ import 'package:mobx_codegen/src/template/observable_stream.dart';
...
+  String get typeName => publicTypeName;
 
-  final Rows<ObservableTemplate> observables = Rows();
-  final Rows<ComputedTemplate> computeds = Rows();
-  final Rows<ActionTemplate> actions = Rows();
-  final Rows<AsyncActionTemplate> asyncActions = Rows();
-  final Rows<ObservableFutureTemplate> observableFutures = Rows();
-  final Rows<ObservableStreamTemplate> observableStreams = Rows();
+  @override
+  String toString() => '''
```
pavanpodila: this comment string should be pulled out as a static constant in `StoreTemplate`

mobx_codegen/pubspec.yaml
pavanpodila: how about `dependency_overrides`

shyndman: I opened #248. Merge that, publish the alpha, and I'll get this changed to reflect it.

mobx_codegen/lib/src/mobx_codegen_base.dart
shyndman: Can you explain more?

Also, do I have to be making sure the user's class implements (or mixes in) Store?

mobx_codegen/lib/src/template/store.dart
shyndman: Done

mobx/lib/src/api/annotations.dart
shyndman: I can't publish the alpha version myself, can I?

mobx_codegen/lib/src/mobx_codegen_base.dart
pavanpodila: We have to treat the `@store` and `Store` mixin as mutually exclusive. However it is possible that someone uses them both together, in which case we should warn them and error out on using only one of the two.

pavanpodila: Also when you use the `@store` annotation, there is no need to use the mixin as we are generating a the public class directly

mobx_codegen/pubspec.yaml
pavanpodila: published `0.3.6-alpha`

pavanpodila: you can use the alpha version now (`0.3.6-alpha`)

mobx_codegen/test/data/valid_annotated_store_output.dart
```diff
@@ -0,0 +1,86 @@
+class User extends _User {
```
pavanpodila: Need to check if `@JsonSerializable` can be used on the private class, since we won't have access to this public one

shyndman: Would you like me to do that, or are you checking?

mobx_codegen/pubspec.yaml
shyndman: 👍 thanks!

mobx_codegen/lib/src/mobx_codegen_base.dart
shyndman: Cool. Got it.

Just out of curiosity, the `Store` class' `dispose()` method. Does anything in the framework call that?

mobx_codegen/test/data/valid_annotated_store_output.dart
shyndman: Don't worry about it. I'll check it.

shyndman: 👍 We're good. `@JsonSerializable` works fine on a private class.

pavanpodila: How do the generated methods look for `@JsonSerializable`? Can they be used from the public facing class instance ? Curious to see some of that code. This is an important use case for many of our users so want to make sure we are not missing it.

shyndman:
OK, let me push up my test project. You can take a look.

mobx_codegen/test/data/valid_annotated_store_output.dart
shyndman: https://github.com/shyndman/private_json_serializable/blob/master/lib/example.dart

mobx_codegen/lib/src/mobx_codegen_base.dart
pavanpodila: No one is calling it. It was added an example to show that you could add such methods and override them in the outer public class.

mobx_codegen/test/data/valid_annotated_store_output.dart
pavanpodila: Nice! 👍

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -20,18 +20,28 @@ import 'package:mobx_codegen/src/template/util.dart';
...
   FutureOr<String> generate(LibraryReader library, BuildStep buildStep) {
-    return library.classes
-        .where(isValidStoreClass)
-        .expand((c) => expandToMixinCode(library, c))
-        .toSet()
-        .join('\n\n');
+    final generatedMixins = library.classes
+        .where(isValidMixinStoreClass)
+        .expand((c) => generateCodeForMixinStore(library, c));
+    final generatedSubclasses = library.classes
```
pavanpodila: Will need a check to ensure only one of `@store` of `Store` mixin is used. We should throw an error and fail the codegen if both are detected on the same class. The pattern we are following is in [errors.dart](https://github.com/mobxjs/mobx.dart/blob/master/mobx_codegen/lib/src/errors.dart)

mobx_codegen/test/data/valid_annotated_store_output.dart
shyndman: Yeah, they're a _little_ strange (lots of underscores), but imo that's fine. This is deep in implementation-land, and I don't think people will mind.

shyndman:
Ready for review :)

mobx_codegen/test/data/valid_annotated_store_output.dart
pavanpodila: I think we should include one test for `@JsonSerializable` as well. This will be useful in the future as well when we discover newer approaches to doing codegen. Please add one.

mobx_codegen/test/mobx_codegen_test.dart
```diff
@@ -37,7 +48,15 @@ void main() {
       const TestInfo(
           description: 'generates for a generic class mixing Store',
           source: './data/valid_generic_store_input.dart',
-          output: './data/valid_generic_store_output.dart')
+          output: './data/valid_generic_store_output.dart'),
+      const TestInfo(
```
pavanpodila: Curious, if you find this way of creating tests good? Are there better approaches?

shyndman: I think this approach is great. Easy to add, and to change over time. I only wish the diff were a little better in the failure case.

mobx_codegen/lib/src/mobx_codegen_base.dart
shyndman: Right, sorry. I got excited. Give me a few.

mobx_codegen/test/mobx_codegen_test.dart
pavanpodila: Ya the failure case is not ideal, only because its a bunch of errors and not in a fixed order. Perhaps we should sort the ordering of error messages in the output 🤔. That should make it easier to create failure output files easily ?

mobx_codegen/test/data/valid_annotated_store_output.dart
shyndman: k

mobx_codegen/test/mobx_codegen_test.dart
shyndman: I actually spent some time debugging one of these test failures. I had to get expected and actual side-by-side to notice the very slight difference I needed to satisfy. I thought to myself, huh, wish it showed me the first difference.

Also, it should be easy to copy the actual string out of the terminal. Right now it's rendered as a string concat in this style:

'this is a line\n'
'this is another line\n'
It'd be nice if it were a multiline, triple-quote literal. Or no quotes at all. That way it's easier to copy and paste.

shyndman: Sorry, brain dump on the test format. Fact is, I do like it. It's a useful way to define tests.

pavanpodila: Ya I have been through that 😄, so I know **exactly** what you underwent!

shyndman:
Hey @pavanpodila, I actually have to step out. I might get a chance to make the requested changes tonight, but if not, tomorrow.

shyndman:
I'm back! Needed rest yesterday. I'm making the requested changes now.

shyndman:
I just realized something about this approach — we're going to have to generate pass-through constructors in the subclass.

Should that be a part of this PR, or a followup?

mobx_codegen/lib/src/mobx_codegen_base.dart
shyndman: Done

shyndman:
Still don't have the JsonSerializable test done, but I did add errors for if people attempt to use mixins and annotations. 

I have to get to work on other stuff, but I'll be back tomorrow for some more. Let me know what you think about the pass-through constructors.

pavanpodila:
Ya we will need the constructor sub

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -20,17 +20,21 @@ import 'package:mobx_codegen/src/template/util.dart';
...
-          .where(isValidStoreClass)
-          .expand((c) => expandToMixinCode(library, c))
-          .toSet()
-          .join('\n\n');
+      _generateCodeForLibrary(library).toSet().join('\n\n');
+
+  Iterable<String> _generateCodeForLibrary(LibraryReader library) sync* {
+    for (final classElement in library.classes) {
+      if (isValidMixinStoreClass(classElement)) {
+        yield* _generateCodeForMixinStore(library, classElement);
```
pavanpodila: What's the benefit of `yield*` instead of `yield` ?

shyndman: `yield*` yields each of the elements of an iterable individually

Without the *, the return type of this method would be `Iterable<Iterable<String>>`

pavanpodila: Ah ok. So it acts as an expansion operator. Cool. Learnt something new!

mobx_codegen/lib/src/errors.dart
```diff
@@ -53,6 +56,26 @@ const fieldPluralizer = Pluralize('the field', 'fields');
...
+      _classNames.add(className);
+    }
+    return condition;
+  }
+
+  @override
+  bool get hasErrors => _classNames.isNotEmpty;
+
+  @override
+  String get message => 'Store classes cannot be defined with the @store '
```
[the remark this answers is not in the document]
shyndman: sg. Done.

shyndman:
Hey @pavanpodila, so the thinking for the `@JsonSerializable` test is that we're going to start including `json_serializable` as a `dev_dependency`? Is that correct?

Will that have the effect of automatically running `json_serializable`'s `Builder` against the users' codebases?

rrousselGit:
> Will that have the side-effect of automatically running json_serializable's Builder against the users' codebases?

That entirely depends on how the generator is configured.

The default behavior is that it runs only on the sources on the package that runs the generator.
But it could be told to run on more than that through the `build.yaml` file.

shyndman:
Excellent. Thanks @rrousselGit.

pavanpodila:
Great, so we can keep the `dev_dependency` on `json_serializable`, right ?

mobx_codegen/lib/src/template/store.dart
```diff
@@ -25,30 +51,27 @@ class StoreTemplate {
...
+  String get actionControllerField => actions.isEmpty
       ? ''
-      : "final $actionControllerName = ActionController(name: '$parentName');";
+      : "final $actionControllerName = ActionController(name: '$parentTypeName');";
 
-  @override
-  // ignore: prefer_single_quotes
-  String toString() => """
-  // ignore_for_file: non_constant_identifier_names, unnecessary_lambdas, prefer_expression_function_bodies, lines_longer_than_80_chars, avoid_as, avoid_annotating_with_dynamic
+  String get storeBody => '''
```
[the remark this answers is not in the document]
pavanpodila: Will have to adjust the output files for tests 😞 but worth it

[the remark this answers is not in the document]
shyndman: Feels like a separate PR, no?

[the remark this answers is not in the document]
pavanpodila: sure lets do that separately 👍

pavanpodila:
If you can move this a branch under the mobx.dart repo, we can get some CI results of testing. Can you push to a branch under here ?

PR #248, opened 2019-08-16
Add the @store annotation

shyndman:
We're splitting this out from the codegen work so that we can get an
alpha package published, and have the codegen package depend on that.

I selected `master` as the merge target, but feel free to change it to 
something more appropriate.


pavanpodila:
Published the alpha: https://pub.dev/packages/mobx

PR #251, opened 2019-08-18
Add info on lack of deep observability

MaikuB:
I also removed a period that i missed in #243 and updated the info about the MobX quick start guide to say nearly all of the concepts are applicable instead of having it say it's 100% applicable given deep observability isn't possible

### DIFF
--- a/docs/src/core-concepts.mdx
+++ b/docs/src/core-concepts.mdx
@@ -75,7 +75,7 @@ will fade away and you will mostly focus on the code within the braces.
 
 The above code is part of the [**Counter Example**](/examples/counter).
 
-For those who have experience with the JavaScript version of MobX, it's important to note that deep observability isn't possible with MobX for Dart as Dart is a statically typed language and there's no support for reflection in Flutter applications. This means with the Dart version of MobX, if a complex object is marked as observable then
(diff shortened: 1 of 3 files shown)


docs/src/core-concepts.mdx
```diff
@@ -75,6 +75,9 @@ will fade away and you will mostly focus on the code within the braces.
 
 The above code is part of the [**Counter Example**](/examples/counter).
 
+For those who have experience with the JavaScript version of MobX, it's important to note that deep observability doesn't apply in the Dart version of MobX. This is because Dart is a statically typed language and there's no support for reflection in Flutter applications. Therefore, if a complex object is marked as observable then
+MobX will not be able to automatically track changes to the object's properties. Should change tracking need to enabled on the object's properties then the properties need to be marked as observable as well.
```
pavanpodila: > Should change tracking need to enabled on the object's properties then the properties need to be marked as observable as well.

I think this sentence should be worded differently.

MaikuB: What's the reason behind your thought and what suggestions do you have? I could change it to "Should change tracking need to enabled on the object's properties then they should be marked as observable as well."

pavanpodila: I was thinking: "Should you need change tracking at the field level, its better to mark them with `@observable`"

MaikuB: done, with a minor tweak to that sentence

PR #252, opened 2019-08-18
Add settings example with ProxyProvider

MaikuB:
- Added an example that for demonstrating using the ProxyProvider and Consumer widgets. Updated relevant docs as well
- Updated code in example to use relative imports (as per https://dart.dev/guides/language/effective-dart/usage#prefer-relative-paths-when-importing-libraries-within-your-own-packages-lib-directory). Can revert if there's a reason I may have missed on why package imports had been used
- Note that I updated the wording in the organizing stores page around the Provider API, as I believe it's more correct to say that it deals with constructing and disposing objects

### DIFF
--- a/docs/src/guides/organizing-stores.mdx
+++ b/docs/src/guides/organizing-stores.mdx
@@ -214,7 +214,7 @@ class MyApp extends StatelessWidget {
 }
 ```
 
-Here, the `ProxyProvider` API allows us to declare that `SettingsStore` depends on an instance of the `PreferencesService` class. When the `SettingsStore` instance is retrieved, it will ensure that its dependencies have been constructed.
+Here, the `ProxyProvider` API allows us to declare that `SettingsStore` depends on an instance of the `PreferencesService` class and will ensure the objects are constructed in the appropriate order based on the chain of dependencies.
 Note that in this particular example, we have used the `Consumer` widget that is an alternative to using `Provider.of<T>` for retrieving a value. Whilst this adds an additional level of indentation, some may prefer this approach as it's
 declarative and assists in building a mental model where a value provided so that can be consumed further down in the widget tree. Furthermore, it facilitates in writing code where dependencies can be more explicit. The following is a widget
 that depends on the `SettingsStore`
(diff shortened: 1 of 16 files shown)


pavanpodila:
I skipped the relative imports only because intellij does auto-imports for me and they package-based imports by default

MaikuB:
Ah ok. Yeah I believe there's an open issue around code completion to suggest relative imports. I'm using an extension for VS Code to get around it at the moment (https://marketplace.visualstudio.com/items?itemName=luanpotter.dart-import). Not sure if there's an equivalent for Intellij

pavanpodila:
I think we should keep package based imports until the tooling supports it. Else the IDE will keep fighting back 😀

MaikuB:
Fair enough. I've made that change now. I feel we should remove `widget_test.dart`. It looks to be the sample widget test for the counter app and as such will fail the tests or was there another reason to keep it?

pavanpodila:
I was planning on adding some widget tests for one of the examples, but may not be necessary. We can check and see if it can be removed.

docs/src/guides/organizing-stores.mdx
```diff
@@ -177,13 +177,98 @@ class CounterListPage extends StatelessWidget {
 }
 ```
 
+For handling when a store depends on a service, the `ProxyProvider` API can be used
```
pavanpodila: I think we should make a separate section here. We can call it "Using ProxyProvider with a service"

mobx_examples/lib/settings/settings_store.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:flutter/foundation.dart';
+import 'package:mobx/mobx.dart';
+import 'preferences_service.dart';
```
pavanpodila: Let's make this a package import too

MaikuB: missed that, thanks

docs/src/guides/organizing-stores.mdx
MaikuB: yep, makes sense

mobx_examples/lib/settings/settings_store.dart
MaikuB: done now. thanks

docs/src/guides/organizing-stores.mdx
MaikuB: done. let me know if further adjustments are needed

PR #256, opened 2019-08-20
Fix type key access ObservableMap

Vardiak:


mobx/lib/src/api/observable_collections/observable_map.dart
```diff
@@ -50,7 +50,7 @@ class ObservableMap<K, V>
       _listenersField != null && _listenersField.hasHandlers;
 
   @override
-  V operator [](Object key) {
+  V operator [](K key) {
```
pavanpodila: This is not a valid override. The base class method has `Object key`, which is why we had to go with `Object`. The only option is to do it inside the method with a type check

pavanpodila:
Closing this for now as there is no traction

PR #258, opened 2019-08-21
Fix change reporting in ObservableMap.[]=

shyndman:
`oldValue` was only being set in certain conditions, which prevented the equality check from having the desired effect.

### DIFF
--- a/mobx/lib/src/api/observable_collections/observable_map.dart
+++ b/mobx/lib/src/api/observable_collections/observable_map.dart
@@ -60,12 +60,11 @@ class ObservableMap<K, V>
   @override
   void operator []=(K key, V value) {
     _context.conditionallyRunInAction(() {
-      V oldValue;
+      final oldValue = _map[key];
       var type = 'set';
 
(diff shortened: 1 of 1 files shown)


mobx/lib/src/api/observable_collections/observable_map.dart
```diff
@@ -60,12 +60,11 @@ class ObservableMap<K, V>
   @override
   void operator []=(K key, V value) {
     _context.conditionallyRunInAction(() {
-      V oldValue;
+      final oldValue = _map[key];
```
pavanpodila: Good find!

pavanpodila: I think you should start making branches on mobx.dart, so we can get the CI tests to detect any potential failures. Can you please submit the PR from a mobx.dart branch? I've sent you the member invite already ;-)

shyndman: Thanks!
Sure.

PR #263, opened 2019-08-22
Flutter mobx forms

pavanpodila:
FormField<T> for simplifying forms development with MobX and Flutter


flutter_mobx_forms/lib/src/form_field.dart
```diff
@@ -0,0 +1,85 @@
...
+  T value;
+
+  @observable
+  bool _isValidating;
+
+  @computed
+  bool get isValidating => _isValidating;
+
+  @computed
+  String get error => errors == null ? null : errors[0];
```
MaikuB: i know it's not ready for review but should the condition here should be `errors == null || errors.isEmpty` so it's more safe?

pavanpodila: No, this is about extracting the first `error` in the list of `errors`, but yes we need to do the `isEmpty` check as well to ensure we don't overflow

flutter_mobx_forms/lib/src/form_field.dart
```diff
@@ -0,0 +1,86 @@
...
+
+  @computed
+  bool get isValid => errors == null;
+
+  SyncFieldValidator<T> validator;
+  AsyncFieldValidator<T> asyncValidator;
+
+  void validate() {
+    if (validator != null && asyncValidator != null) {
+      throw ArgumentError(
```
rrousselGit: What about having `validator`/`asyncValidator` as `final`, and move this as assert on the constructor?

pavanpodila: Yup that's a good idea. Will do

flutter_mobx_forms/lib/src/form_field.dart
```diff
@@ -0,0 +1,92 @@
...
+abstract class _FormField<T> with Store {
+  _FormField(
+      {this.name,
+      this.label,
+      this.value,
+      this.validator,
+      this.asyncValidator,
+      this.validationPolicy = ValidationPolicy.manual}) {
+    // ignore: prefer_asserts_in_initializer_lists
+    assert(validator != null && asyncValidator != null,
```
rrousselGit: can't it be:

```dart
  _FormField({this.name,
      this.label,
      this.value,
      this.validator,
      this.asyncValidator,
      this.validationPolicy = ValidationPolicy.manual,
  }): assert(validator != null && asyncValidator != null, 'Only one of validator or asyncValidator can be specified');
```

?

pavanpodila: Definitely could, just felt like a really loaded constructor line :-). If that's the idiomatic approach, I'm happy to change. .... actually changed

flutter_mobx_forms/lib/src/form_field.dart
```diff
@@ -0,0 +1,92 @@
...
+      errors == null ? null : (errors.isEmpty ? null : errors[0]);
+
+  @computed
+  bool get isValid => errors == null;
+
+  final SyncFieldValidator<T> validator;
+  final AsyncFieldValidator<T> asyncValidator;
+
+  void validate() {
+    if (validator != null) {
```
rrousselGit: Should we prevent validation inside validation?

```suggestion
    assert(!_isValidating);
    if (validator != null) {
```

flutter_mobx_forms/lib/src/form_field.dart
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @action
+  void _syncValidate() {
+    try {
+      _isValidating = false;
+
+      final errors = validator(value);
+      this.errors = errors == null ? null : errors.toList(growable: false);
+    } on Object catch (_) {
```
rrousselGit: This block can be removed entirely
That's the default behavior of `try {} finally {}` without a `catch`

pavanpodila: Ah nice! didn't know that. Thanks for the tip!

rrousselGit:
I'm still convinced that `error` should be read-only/computed by the validator, instead of manually applied.

Random thought: What about making error's type generic?

```dart
class FormField<Value, Error> {
  @observable
  Value value;


  Error get error;
}
```

In the same fashion, `asyncValidator` could be a factory constructor:

```dart
class FormField<A, Error> {
  FormField({Error validator(A value)});
  factory FormField.asyncValidator({Future<Error> validator(A value)}) = _AsyncFormField;
}
```

where `_AsyncFormField` is a subclass of `FormField` with a reaction on the result of `validator`, to update `error` accordingly.

pavanpodila:
`error` is a computed right now, so its already read-only. The only addition I made is to not limit to only one error, instead you can have an array. I can make the `errors` read-only and have the validator update a private field.

flutter_mobx_forms/lib/src/form_field.dart
```diff
@@ -0,0 +1,115 @@
...
+part 'form_field.g.dart';
+
+typedef SyncFieldValidator<T> = Iterable<String> Function(T value);
+typedef AsyncFieldValidator<T> = Stream<String> Function(T value);
+
+enum ValidationPolicy { onChange, always, manual }
+
+class FormField<T> = _FormField<T> with _$FormField;
+
+abstract class _FormField<T> with Store {
```
Solido: Flutter already use FormField as class name can we have something different to better handle friction when integration our Form with Flutter widget ?

flutter_mobx_forms/lib/src/form_field.dart
```diff
@@ -0,0 +1,115 @@
...
+    } finally {
+      _isValidating = false;
+    }
+  }
+
+  @action
+  // ignore: avoid_void_async
+  Future<void> _asyncValidate() async {
+    try {
+      _isValidating = false;
```
Solido: Shouldn't this be true ? idem for sync

pavanpodila: Oops yes!

flutter_mobx_forms/lib/src/form_field.dart
pavanpodila: Suggestions? MobxFormField?

rrousselGit: >  MobxFormField?

I don't like it. This kind of name is acheivable using alias imports:

```dart
import 'package:flutter_mobx_form/flutter_mobx_form.dart' as mobx;

mobx.FormField foo;
```


What about:

- AutoFormField
- FieldState
- ObserverFormField

pavanpodila: I like `FieldState` better. It reflects the intent more clearly. Aliasing is another option but then can we use other identifiers without the alias? as in `@observable` instead of `@mobx.observabble`

pavanpodila: `FormField` is now `FieldState`

rrousselGit: Yes since `observable` is from another package

pavanpodila: ah nice, that's good to know. That said, I still feel `FieldState` captures the intent more clearly, we can also have `FieldGroup` for a set of related fields. Ultimately, a Form is just a `FieldGroup` at the root level. I am borrowing this concept from Angular, where I first saw it.

Solido: Interesting so we can have reactions at a group level based on value of one of its fields or other group ?

pavanpodila: Yup, we can rollup errors / values all the way to the top

flutter_mobx_forms/lib/src/field_state.dart
```diff
@@ -0,0 +1,111 @@
...
+
+  @computed
+  bool get isValid => errorContext.isEmpty;
+
+  final SyncFieldValidator<T> validator;
+  final AsyncFieldValidator<T> asyncValidator;
+
+  void validate() {
+    assert(!_isValidating);
+
```
Solido: Missing  `errorContext.reset();`

aloisdeniel:
What is the best way to manage computed values ?

Thought, the following example may be misunderstood by newcomers. 

```dart
@bindable
FieldState<String> firstname = FieldState<String>(...)

@bindable
FieldState<String> lastname = FieldState<String>(...)

@computed
int fullname => "$firstname $lastname";
```

pavanpodila:
What do you mean by "manage computed values"?

aloisdeniel:
In my previous example, does the the `fullname` raise an update whenever `firstname.value` changes ?

pavanpodila:
Yes it does. It behaves like a read-only Observable that notifies on changes

rrousselGit:
By the way, computed properties are cached right?
If we request it twice but no Observer it depends on changed, then the getter isn't called again is it?

pavanpodila:
The caching is not on the value but the notification. Calling it twice will recompute the value but send out the notification only of there is a change.

pavanpodila:
Will pick this up later

PR #269, opened 2019-08-22
make main function in examples return a future

MaikuB:
Was looking at an example app that one of the members of the Flutter team and found out that the main function can be made to return a `Future`. This should allow us to having an async void function and remove the need to suppress the warning


pavanpodila:
Nice, didn't know that. Great tip @MaikuB

PR #273, opened 2019-08-25
Add support for codegen of "unknown" type names

### LINKED ISSUE #165: mobx_codegen generates invalid code when observable is of type Color
I'm trying to define a model which has attributes of type Color (dart.ui.Color):

```
part "Label.g.dart";

class Label = LabelBase with _$Label;
abstract class LabelBase  implements Store {
  @observable 
  Color bgColor;
}
```

The generated code in Label.g.dart, however, uses `dynamic`:

  final _$bgColorAtom = Atom(name: 'LabelBase.bgColor');

```
  @override
  dynamic get bgColor {
    _$bgColorAtom.reportObserved();
    return super.bgColor;
  }

  @override
  set bgColor(dynamic value) {
    _$bgColorAtom.context.checkIfStateModificationsAreAllowed(_$bgColorAtom);
    super.bgColor = value;
    _$bgColorAtom.reportChanged();
  }
```

which results in a compile-time error:

> Error: The return type of the method '_$Label.bgColor' is 'dynamic', which does not match the return type, 'Color', of the overridden method, 'LabelBase.bgColor'.
>  - 'Color' is from 'dart:ui'.                                           
> Change to a subtype of 'Color'.                                         
>   dynamic get bgColor {            

Am I doing something wrong or is it a problem with mobx_codegen?

### LINKED ISSUE #267: Generator @store constructors are missing parameter type arguments
```dart
@store
class _User {
  _User(Iterable<int> ints);
}
```
generates
```dart
class User extends _User {
  User(Iterable ints) : super(ints);
}
```

Note that `<int>` is missing from the parameter's type annotation.

### LINKED ISSUE #268: @store annotated classes cannot reference other @stores
If a `@store` references other stores in its `@observable`s, return types, or parameter types, their types are not recognized and get generated as dynamic.

For example
```dart
@store
class _Car {
  @observable
  Engine engine;
}

@store 
class _Engine {
  @observable
  int pistonCount;
}
```
is generated as
```dart
class Car extends _Car {
  dynamic get engine { ... }
  set engine(dynamic value) { ... }
}
```
Which makes sense — the analyzer can't analyze a type before it's been generated.

That said, we can get around this by looking directly at the AST, and using the type name the user wrote.

_Unfortunately_, the `build` package doesn't make this easy on us. `CompilationUnit` (the AST root) isn't exposed by the `Element` in this environment...but that doesn't mean we can't re-compute it.

In essence, the solution might be something like this (for the example above):
```dart
// element is the FieldElement

// Parse the file, without resolving types (we'd cache this tree)
final result = parseFile2(
  path: element.source.fullName,
  resourceProvider: element.session.resourceProvider,
  featureSet: FeatureSet.fromEnableFlags([]),
);

// Find the AST node belonging to the element we're evaluating. If it's a field, 
// this will find a SimpleIdentifier node.
final nameNode = NodeLocator(element.nameOffset).searchWithin(result.unit);

// Walk up the ancestry to find the node that contains type information
final variableNode =
    nameNode.thisOrAncestorMatching((node) => node is VariableDeclarationList);

// Grab the type, as the user wrote it. If it works, it works, if it doesn't, it 
// doesn't. Their problem.
variableNode.type.toString(); // "Engine"
```

It won't be a ton of code, and I'd be surprised if there was a noticeable performance difference (parsing is cheap). But it's your call whether or not it's worth it.

Oh, and notably, this will resolve problems encountered with `dart:ui` types as well.

shyndman:
If the analyzer is unable to resolve a type, it reports that the type is
"dynamic", and our codegen writes it out way. This situation is
particularly noticeable because the build package's analyzer draws from
a pre-computed analysis summary that does not include dart:ui.

The recent work on annotated stores demonstrates a similar and related
problem, where we'd like to reference generated types in our store
definitions, but because the analyzer doesn't know about types that have
not yet been generated, dynamics litter the generated code.

This PR explores an approach where we output types to generated code
as they were written in the input instead of relying on the analyzer's
interpretation. If we encounter a type that the analyzer reports as
dynamic, we look up the element's AST node, and use the name found
there. Magic.

Fixes #268 
Fixes #267 
Fixes #165

### DIFF
--- a/mobx_codegen/lib/src/template/method_override.dart
+++ b/mobx_codegen/lib/src/template/method_override.dart
@@ -1,8 +1,8 @@
 import 'package:analyzer/dart/element/element.dart';
-import 'package:analyzer/dart/element/type.dart';
 import 'package:mobx_codegen/src/template/comma_list.dart';
 import 'package:mobx_codegen/src/template/params.dart';
 import 'package:mobx_codegen/src/template/util.dart';
+import 'package:mobx_codegen/src/type_names.dart';
 
 /// Stores templating information about constructors and methods.
 class MethodOverrideTemplate {
@@ -12,7 +12,7 @@ class MethodOverrideTemplate {
     // ignore: prefer_function_declarations_over_variables
     final param = (ParameterElement elem) => ParamTemplate()
       ..name = elem.name
-      ..type = elem.type.displayName
+      ..type = findParameterTypeName(elem)
       ..defaultValue = elem.defaultValueCode;
 
     final positionalParams = method.parameters
@@ -25,19 +25,15 @@ class MethodOverrideTemplate {
     final namedParams =
         method.parameters.where((param) => param.isNamed).toList();
 
-    final returnType = method.returnType;
-    final returnTypeArgs = returnType is ParameterizedType
-        ? returnType.typeArguments.map((p) => p.displayName).toList()
-        : <String>[];
-
     this
       ..name = method.name
-      ..returnType = method.returnType.displayName
+      ..returnType = findReturnTypeName(method)
       ..setTypeParams(method.typeParameters.map(typeParamTemplate))
       ..positionalParams = positionalParams.map(param)
       ..optionalParams = optionalParams.map(param)
       ..namedParams = namedParams.map(param)
-      ..returnTypeArgs = SurroundedCommaList('<', '>', returnTypeArgs);
+      ..returnTypeArgs = SurroundedCommaList(
+          '<', '>', findReturnTypeArgumentTypeNames(method));
   }
 
   String name;
--- a/mobx_codegen/test/data/valid_annotated_store_referencing_store_input.dart
+++ b/mobx_codegen/test/data/valid_annotated_store_referencing_store_input.dart
@@ -1,0 +1,50 @@
+library generator_sample;
+
+import 'dart:ui' as ui;
+
+import 'package:mobx/mobx.dart';
+
+part 'generator_sample.g.dart';
+
+@store
+class _Car {
+  _Car(this.engine);
+
+  @observable
+  ui.Color paintColor;
+
+  @observable
+  Engine engine;
+
+  @observable
+  ObservableList<Tire> tires = ObservableList<Tire>();
+
+  @observable
+  Windshield windshield = Windshield();
+
+  @action
+  Future<List<Tire>> changeTiresIfRequired() async => [];
+
+  @action
+  Future<List<T>> changeCarPartsIfRequired<T extends CarPart>() async => [];
+}
+
+@store
+class _CarPart {}
+
+@store
+class _Engine extends CarPart {}
+
+@store
+class _Tire extends CarPart {}
+
+@store
+class _Windshield extends CarPart {
+  @observable
+  Stream<List<Bug>> squashedBugs() async* {
+    yield <Bug>[];
+  }
+}
+
+@store
+class _Bug {}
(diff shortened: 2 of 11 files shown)


mobx_codegen/lib/src/template/method_override.dart
```diff
@@ -1,8 +1,9 @@
 import 'package:analyzer/dart/element/element.dart';
-import 'package:analyzer/dart/element/type.dart';
-import 'package:mobx_codegen/src/template/comma_list.dart';
-import 'package:mobx_codegen/src/template/params.dart';
-import 'package:mobx_codegen/src/template/util.dart';
+
+import '../type_names.dart';
```
shyndman: Oops, a tool I used relativized these. I can revert if you'd like.

pavanpodila: Ya let's keep as package imports for consistency

shyndman: Done

shyndman:
Ah, I see the coverage failures. Let me get that number up.

mobx_codegen/test/data/valid_store_with_ui_types_input.dart
```diff
@@ -0,0 +1,36 @@
+library generator_sample;
+
+import 'dart:ui';
+
+import 'package:mobx/mobx.dart';
+
+part 'generator_sample.g.dart';
```
pavanpodila: I think we can remove this line and the test should still work. Can we try that.

mobx_codegen/test/data/valid_annotated_store_referencing_store_input.dart
```diff
@@ -0,0 +1,50 @@
+library generator_sample;
+
+import 'dart:ui' as ui;
+
+import 'package:mobx/mobx.dart';
+
+part 'generator_sample.g.dart';
```
pavanpodila: I think we can remove the part file declarations from all input files. Tests should still work I think

pavanpodila:
Looks like the code coverage has dropped from 100%. We should include the examples for the red lines here: https://codecov.io/gh/mobxjs/mobx.dart/pull/273/diff?src=pr&el=tree#diff-bW9ieF9jb2RlZ2VuL2xpYi9zcmMvdHlwZV9uYW1lcy5kYXJ0

pavanpodila:
Some failing tests in `mobx_codegen`

shyndman:
> Some failing tests in `mobx_codegen`

Looks like a flake, which I've seen a bunch of. All tests pass, but something else is failing, oddly. Take a look at the logs.

shyndman:
Alright, there we go. Code coverage is back up.

pavanpodila:
> Alright, there we go. Code coverage is back up.

Great. Once we are through this, I want to also investigate how to leverage CircleCI better. We may not be using to the best of its potential ?

shyndman:
It strikes me as strange that external PRs don't run through CI. There has
to be a way to fix that.

mobx_codegen/test/data/valid_annotated_store_referencing_store_input.dart
shyndman: As it turns out, we cannot. In their absence, generator is reported to be an empty string by the tests.

pavanpodila:
It's because the CI pipeline is defined at the Repo level and for its branches. We may be able to change that for external PRs, will have to investigate

shyndman:
https://circleci.com/docs/2.0/oss/

mobx_codegen/test/data/valid_store_with_ui_types_input.dart
shyndman: Addressed in previous comment.

pavanpodila:
image
Done!

shyndman:
Nice!

How's the PR look?

PR #275, opened 2019-08-25
Include a set reference in ObservableSet change notifications

shyndman:
This mimics the other Observable collections.

https://github.com/mobxjs/mobx.dart/blob/master/mobx/lib/src/api/observable_collections/observable_list.dart#L380
https://github.com/mobxjs/mobx.dart/blob/master/mobx/lib/src/api/observable_collections/observable_map.dart#L210

### DIFF
--- a/mobx/lib/src/api/annotations.dart
+++ b/mobx/lib/src/api/annotations.dart
@@ -13,6 +13,17 @@ class MakeObservable {
 /// documentation
 const MakeObservable observable = MakeObservable._();
 
+class MakePanoptic {
+  const MakePanoptic._();
+}
+
+/// Declares a class field as an _actual_ observable.
+///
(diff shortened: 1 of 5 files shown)


pavanpodila:
Coverage has dropped a bit

shyndman:
So I'm looking at the diff...how could coverage have dropped? There are no new branches.

shyndman:
Oh, I see, it's the `asserts` in the initializer list. That seems alright, doesn't it?

pavanpodila:
How about a quick test just on the `SetChange` class without providing the required values. That would throw the `AssertionError` and we have full coverage?

shyndman:
Done

shyndman:
Oops, still figuring out how to use GH the way I'd like. Kicking off a new PR.

shyndman:
Continued in #276

PR #276, opened 2019-08-25
Include a set reference in ObservableSet change notifications

shyndman:
This mimics the other Observable collections.


shyndman:
This includes tests for null args.

PR #282, opened 2019-08-27
_isValidating can't be null

jascodes:
Because assert(!_isValidating) is asserting negate of null which is illegal.


pavanpodila:
Thanks @jascodes!

PR #283, opened 2019-08-28
close off brackets correctly  in organising stores page

MaikuB:
Missed that I incorrectly used the square brackets for part of the text

### DIFF
--- a/docs/src/guides/organizing-stores.mdx
+++ b/docs/src/guides/organizing-stores.mdx
@@ -244,7 +244,7 @@ class SettingsExample extends StatelessWidget {
 }
 ```
 
-Using a non-default constructor here allows us to explicit declare that the `SettingsExample` class depends on the `SettingsStore` for it to function. Some developers may already be familiar with structuring code this way
+Using a non-default constructor here allows us to explicitly declare that the `SettingsExample` class depends on the `SettingsStore` for it to function. Some developers may already be familiar with structuring code this way
 when working with applications/frameworks that support constructor injection. An instance of the `SettingsStore` is passed through the `Consumer` widget as per the example code below
 
 ```dart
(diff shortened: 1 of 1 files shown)


docs/src/guides/organizing-stores.mdx
```diff
@@ -260,7 +260,7 @@ Example(
 ...
 ```
 
-A side effect of this is that writing widget tests become easier as well as we'd know what dependencies a widget has. We could then pass, say, mocks (e.g. using the [mockito][2] package] of said dependencies can be passed if required.
+A side effect of this is that writing widget tests become easier as well as we'd know what dependencies a widget has. We could then pass, say, mocks (e.g. using the [mockito][2] package) of said dependencies can be passed if required.
```
pavanpodila: Can be changed like so for a simpler language:

```
A side effect of this is that writing widget tests become easier, as the dependencies are well known. We could then pass mocks (e.g. using the [mockito][2] package) of said dependencies, if required.
```

MaikuB: Thanks. Reworded it a bit now

MaikuB:
@pavanpodila one thing I've noticed is that the references don't seem to be appearing on the bottom anymore. I'm not sure what's changed there so may need your help with that

docs/src/guides/organizing-stores.mdx
```diff
@@ -260,7 +260,7 @@ Example(
 ...
 ```
 
-A side effect of this is that writing widget tests become easier as well as we'd know what dependencies a widget has. We could then pass, say, mocks (e.g. using the [mockito][2] package] of said dependencies can be passed if required.
+A side effect of this is that writing widget tests become easier as well. Since the widget's dependencies have been made explicit, we could pass mocks (e.g. using the [mockito][2] package) of said dependencies if required.
```
pavanpodila: a comma is needed after "said dependencies"

pavanpodila:
> @pavanpodila one thing I've noticed is that the references don't seem to be appearing on the bottom anymore. I'm not sure what's changed there so may need your help with that

I see..will take a look after we merge

docs/src/guides/organizing-stores.mdx
MaikuB: It's not said with a pause to warrant a comma. A similar usage would be, say, "you can call me if needed"

pavanpodila: got it

PR #285, opened 2019-08-29
Report errors that occur during Observer.setState

shyndman:
Errors that occur internal to the setState() call are now reported via FlutterError.reportError so they don't go unnoticed by the user.

### DIFF
--- a/flutter_mobx/lib/src/observer.dart
+++ b/flutter_mobx/lib/src/observer.dart
@@ -29,31 +29,41 @@ class Observer extends StatefulWidget {
   final WidgetBuilder builder;
 
   @visibleForTesting
-  Reaction createReaction(Function() onInvalidate) {
+  Reaction createReaction(
+    Function() onInvalidate, {
+    Function(Object, Reaction) onError,
+  }) {
     final ctx = context ?? mainContext;
 
-    return ReactionImpl(ctx, onInvalidate, name: name);
+    return ReactionImpl(ctx, onInvalidate, name: name, onError: onError);
   }
 
   @override
-  State<Observer> createState() => _ObserverState();
+  State<Observer> createState() => ObserverState();
 
   void log(String msg) {
     debugPrint(msg);
   }
 }
 
-class _ObserverState extends State<Observer> {
+@visibleForTesting
+class ObserverState extends State<Observer> {
   ReactionImpl _reaction;
 
   @override
   void initState() {
     super.initState();
 
-    _reaction = widget.createReaction(_invalidate);
+    _reaction = widget.createReaction(invalidate, onError: (e, _) {
+      FlutterError.reportError(FlutterErrorDetails(
+        library: 'flutter_mobx',
+        exception: e,
+        stack: e is FlutterError ? e.stackTrace : null,
+      ));
+    });
   }
 
-  void _invalidate() => setState(noOp);
+  void invalidate() => setState(noOp);
 
   static void noOp() {}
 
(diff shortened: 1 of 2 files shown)


flutter_mobx/lib/src/observer.dart
```diff
@@ -50,7 +53,14 @@ class _ObserverState extends State<Observer> {
   void initState() {
     super.initState();
 
-    _reaction = widget.createReaction(_invalidate);
+    _reaction = widget.createReaction(_invalidate, onError: (e, _) {
```
pavanpodila: Need a test as well for this

flutter_mobx/lib/src/observer.dart
```diff
@@ -50,7 +53,14 @@ class _ObserverState extends State<Observer> {
...
+      // If we're receiving a FlutterError, it's occurring internally in the
+      // setState() call. As such, we throw the exception so that it can be
+      // routed through Flutter's error handling machinery.
+      if (e is FlutterError) {
```
pavanpodila: Why only FlutterError? Why not Error?

rrousselGit: Why not:

```suggestion
    FlutterError.reportError(FlutterErrorDetails(
      library: 'mobx',
      exception: e,
      stack: stacktrace,
    ));
```

shyndman: Nice! We'll probably want that original stack if we've got it though.

shyndman:
Also, I'm in the process of moving back to Canada, so will be afk until Monday.

flutter_mobx/lib/src/observer.dart
shyndman: Done.

flutter_mobx/lib/src/observer.dart
shyndman: Suggestion accepted. Thanks.

shyndman:
Comments addressed. PTAL.

flutter_mobx/lib/src/observer.dart
```diff
@@ -29,31 +29,41 @@ class Observer extends StatefulWidget {
...
   ReactionImpl _reaction;
 
   @override
   void initState() {
     super.initState();
 
-    _reaction = widget.createReaction(_invalidate);
+    _reaction = widget.createReaction(invalidate, onError: (e, _) {
+      FlutterError.reportError(FlutterErrorDetails(
+        library: 'mobx',
```
pavanpodila: library should be `flutter_mobx`

flutter_mobx/test/flutter_mobx_test.dart
```diff
@@ -35,6 +39,23 @@ class LoggingObserver extends Observer {
...
+    Key key,
+  }) : super(key: key, builder: builder);
+
+  @override
+  State<Observer> createState() => FlutterErrorThrowingObserverState();
+}
+
+class FlutterErrorThrowingObserverState extends ObserverState {
+  @override
+  void invalidate() => throw FlutterError('setState failed!');
```
pavanpodila: Curious what will happen if the error is just a regular `Error` (eg: `AssertionError`). I think we should have one test for generic errors.

shyndman:
Done. Added two additional tests, for a non-Flutter error and an exception. Changed library name. PTAL.

shyndman:
Also, after merge, mind if we push a new version to pub? This is a big fix for us.

pavanpodila:
> Also, after merge, mind if we push a new version to pub? This is a big fix for us.

Done: https://pub.dev/packages/flutter_mobx

shyndman:
Thanks Pavan :)

PR #287, opened 2019-09-05
Fix a case where an unknown parameter type is missed

shyndman:
Turns out that a `NormalFormalParameter` is not always the node that we find when searching for a `ParameterElement`'s counterpart in the AST. If a `DefaultFormalParameter` is discovered, we will now traverse to the `SimpleFormalParameter` containing the type name.

### DIFF
--- a/mobx_codegen/lib/src/type_names.dart
+++ b/mobx_codegen/lib/src/type_names.dart
@@ -32,7 +32,9 @@ String findParameterTypeName(ParameterElement parameter) {
     return _findInitializingParameterTypeName(parameter);
   }
 
-  final node = _findElementNode<NormalFormalParameter>(parameter);
+  final node =
+      _findElementNode<DefaultFormalParameter>(parameter)?.parameter ??
+      _findElementNode<NormalFormalParameter>(parameter);
   if (node is SimpleFormalParameter) {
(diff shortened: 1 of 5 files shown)


shyndman:
The code coverage is misleading. In terms of functionality, we're just as tested. I'm not sure how to bring it up (but imho, we shouldn't bother).

pavanpodila:
> The code coverage is misleading. In terms of functionality, we're just as tested. I'm not sure how to bring it up (but imho, we shouldn't bother).
> 
> > The diff coverage is 100%.
> 
> > Absolute coverage decreased by -0.23% but relative coverage remained the same compared to [c1e1d4b](https://github.com/mobxjs/mobx.dart/commit/c1e1d4b40ad890efe86e101d4f78676ef07ecdc7)

Ya not a stickler for coverage as long as the functionality is tested and we are confident about it

shyndman:
Excellent. Any chance we can get a point release? I'm going to start moving over our application to `@store`, since it sounds like it's going to stay.

pavanpodila:
Yup will do in few minutes

pavanpodila:
done: 0.3.8

shyndman:
thank you sir :)

Issue #249, opened 2019-08-17
Update API docs to include all the options for reactions


pavanpodila:
The API docs for reactions do not show the various options available for `autorun`, `reaction`, `when`, `asynWhen`. This needs to be added.

Ashutosh-cm:
Hi! I'd like to work on this issue if it's still available

AzazelSensei:
Opened https://github.com/mobxjs/mobx.dart/pull/1085 — documents the optional args on `when` and `asyncWhen`.

PR #295, opened 2019-09-17
Expose versions

### LINKED ISSUE #213: Add version strings to all packages
At runtime we don't have a way to know the versions of `mobx`, `mobx_codegen` and `flutter_mobx`. Need to add the version strings for all these packages.

Something like:

```dart
const String version = 'x.y.z';
```

Also need to update release guidelines in the docs project

esarbanis:
This is an attempt to automate the version synchronization between the pubspecs and the library files.

Closes #213

### DIFF
--- a/docs/src/development/release.mdx
+++ b/docs/src/development/release.mdx
@@ -17,7 +17,15 @@ If you are a developer on the MobX project, here are few guidelines that will he
 version: 0.2.0
 ```
 
-### 2. Changelog
+### 2. Expose version
+
+- Expose the updated version of `pubspec.yaml` to the library `.dart` file by running the appropriate command:
+
+```bash
+./tool/expose_version.dart <package_name>
+```
+
+### 3. Changelog
 
 - Update the `CHANGELOG.md` file to include the changes introduced in the new version. Normally you would do this by adding a new bullet point for the version and follow the template from earlier versions.
 - If the changes are minimal, you can also append the change to an earlier bullet point but include the new version as a range. For example, the changes in `0.3.2+3` were minimal, so we added it as a range from the earlier version.
@@ -26,7 +34,7 @@ version: 0.2.0
 ## 0.3.0 - 0.3.2+3
 ```
 
-### 3. Publish
+### 4. Publish
 
 Run the command `flutter packages pub publish` in the package folder. Assuming you have done all of the earlier steps, this should be just a matter of hitting the `y` key and publishing.
 
--- a/mobx/lib/mobx.dart
+++ b/mobx/lib/mobx.dart
@@ -61,3 +61,5 @@ export 'package:mobx/src/core.dart'
         NotificationHandlers,
         ReactionImpl,
         DerivationState;
+
+const version = '0.3.7';
--- a/tool/expose_version.dart
+++ b/tool/expose_version.dart
@@ -1,0 +1,28 @@
+#!/usr/bin/env dart
+import 'dart:io';
+
+void main(List<String> packages) {
+  for (var package in packages) {
+    final libraryFile = File('./$package/lib/$package.dart');
+    final content = libraryFile.readAsStringSync();
+    final strippedContent = stripVersion(content);
+
+    final pubspecFile = File('./$package/pubspec.yaml');
+    final projectData = pubspecFile.readAsStringSync();
+    final versionNumber = resolveVersion(projectData);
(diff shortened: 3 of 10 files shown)


mobx/lib/mobx.dart
```diff
@@ -61,3 +61,4 @@ export 'package:mobx/src/core.dart'
         NotificationHandlers,
         ReactionImpl,
         DerivationState;
+const version='0.3.7';
```
pavanpodila: Should probably do this as the top line after the `library` statement ? Easier to spot then. Same for all other packages

tool/expose_version.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
pavanpodila: Can we make this script more readable by using dart / python ? Would make it easier for maintainability :-)

pavanpodila: Great direction and thought process!

esarbanis:
🤔 I don't quite know how to approach this reduced coverage. Should I create a test just for the versions?

pavanpodila:
> 🤔 I don't quite know how to approach this reduced coverage. Should I create a test just for the versions?

Ya just add a test to check the version string is exported.

tool/expose_version.sh
esarbanis: Thanks @pavanpodila !

I didn't know the way you would have liked to go, that's why I created it in "simple" bash.

I will attempt a dart port if you are ok with it.

pavanpodila: Your bash is way better than mine :-). Didn't look simple at all.

Ya dart would be good!

mobx/lib/mobx.dart
esarbanis: It was easier to append to the file, because of the bash script 😄 . I plea guilty for being lazy 😉

pavanpodila: you got something working, which is a great step forward! The suggestion is just a tiny improvement to what you have done :-)

esarbanis: Hm, I cannot actually put it anywhere else but the bottom of the file because of `Directives must appear before any declaration`.

tool/expose_version.dart
```diff
@@ -0,0 +1,28 @@
...
+
+String createVersionVariable(String versionNumber) =>
+    "const version = '$versionNumber';";
+
+String resolveVersion(String projectData) =>
+    RegExp('version: ([0-9]*\.[0-9]*\.[0-9]*)')
+        .firstMatch(projectData)
+        .group(1);
+
+String stripVersion(String content) => content.replaceAll(
```
pavanpodila: The regex should also include the pre release version strings

esarbanis: I removed the complexity of verifying the smever all together.

docs/src/development/release.mdx
```diff
@@ -17,7 +17,15 @@ If you are a developer on the MobX project, here are few guidelines that will he
 version: 0.2.0
 ```
 
-### 2. Changelog
+### 2. Expose version
+
+- Expose the updated version of `pubspec.yaml` to the library `.dart` file by running the appropriate command:
+
+```bash
+./tool/expose_version.dart <package_name>
```
pavanpodila: Since we only have 3 packages (as of now), we could just make this command as 

```sh
./tool/expose_version.dart mobx flutter_mobx mobx_codegen
```

esarbanis: This part of the instructions is under the context of a single package, that's why I mentioned it like that. Otherwise, one should complete the first step of all packages and then run the multi-command.

The idea is that the changes of `expose_version.dart` should be included in the commit of the version bump and must run after the version update in `pubspec.yaml`.

Do I miss something?

pavanpodila: Yup sounds good!

pavanpodila:
Thanks @esarbanis. This is an awesome change that greatly simplifies our life!

esarbanis:
I am glad to have helped 😄 is there anything else I might start working on?

pavanpodila:
Thanks for the offer :-). One thing that would need a fresh pair of eyes is looking at the docs on https://mobx.pub. There are few areas to improve like documenting the API more exhaustively and giving a general proof read to ensure it sound meaningful for novice to expert alike.

I have an open issue around that: #249 but the scope for that can be expanded

PR #305, opened 2019-10-13
Add ObservableValue class

### LINKED ISSUE #301: [Proposal] Observable value interface
Suppose we have a service that returns the current user's geolocation. The geolocation changes over time, so it should be observable. The question is what type the service should return.
It can be a raw value:
```dart
abstract class GeoLocationService {
    GeoLocation get geoLocation;
}
```
The getter will resolve an observable value under the hood, as a result geoLocation consumers will be notified.
I don't like this approach. It is not clear from the service interface that the property is observable.

Another way is to return an observable value:
```dart
abstract class GeoLocationService {
    Observable<GeoLocation> get geoLocation;
}
```
This is a bad solution because it allows to modify the property from the outside:
`service.geoLocation.value = newValue;`
Moreover, it imposes an implementation restriction - `Observable` cannot be replaced by `ObservableStream` or `Computed` without breaking the interface.

I would prefer to return a simple interface `ObservableValue` (actually, it would be better to call it `Observable`, but this name is already in use 😞)
```dart
abstract class ObservableValue<T> {
    T get value;
}
```
which will be implemented by `Observable`, `Computed`, `ObservableStream`, `ObservableFuture`.

The benefit of such an interface is even more significant when we need to return observable values from functions and process them later:
```dart
abstract class GeoLocationSharingService {
    ObservableValue<GeoLocation> requestGeoLocation(String userName);
}

final myLocation = geoLocationService.geoLocation;
final myFriendsLocations = ObservableList<ObservableValue<GeoLocation>>();
myFriendsLocations.add(geoLocationSharingService.requestGeoLocation('Artur'));
myFriendsLocations.add(geoLocationSharingService.requestGeoLocation('Maria'));
final distanceToNearestFriend = Computed<double>(() => nearestDistance(myFriendsLocations, myLocation));
```

Update: PR https://github.com/mobxjs/mobx.dart/pull/305

t-artikov:
This allows to work with observable classes(`Observable`, `Computed`, `ObservableFuture`, `ObservableStream`) through a common interface.

Fixes #301

### DIFF
--- a/mobx/lib/src/api/async/observable_future.dart
+++ b/mobx/lib/src/api/async/observable_future.dart
@@ -41,7 +41,7 @@ class FutureResult<T> {
   }
 }
 
-class ObservableFuture<T> implements Future<T> {
+class ObservableFuture<T> implements Future<T>, ObservableValue<T> {
   /// Create a new observable future that tracks the state of the provided future.
   ObservableFuture(Future<T> future, {ReactiveContext context})
       : this._(context ?? mainContext, future, FutureStatus.pending, null);
@@ -91,6 +91,7 @@ class ObservableFuture<T> implements Future<T> {
   /// Value if this completed with a value.
   ///
   /// Null otherwise.
+  @override
   T get value => status == FutureStatus.fulfilled ? _result.result : null;
 
   /// Error value if this completed with an error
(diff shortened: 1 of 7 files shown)


pavanpodila:
Looks good overall. Need some fixes in the tests for the mobx_codegen and mobx packages. Good to merge after that

t-artikov:
Analyzer warnings fixed now.

PR #309, opened 2019-10-22
Upgrade the analyzer to support a range of versions

### LINKED ISSUE #130: importing with alias causing partial to not generate correctly
importing an observable type like this:
```dart
import 'package:example_flutter/models/event.dart' as model;
```

and declaring an observable:
```dart
@observable
ObservableList<model.Event> events = ObservableList<model.Event>();
```

causes the `.g` file to not be compiled correctly.

shyndman:
Allow a version range for the analyzer package so that downstream packages can take advantage of the newer build_resolver package. Of note, the newer build_resolver resolves dart:ui types.

This also changes the supertype check for mixin-style stores to use the TypeSystem class. The previously used method is deprecated in the latest analyzer.

Updating the analyzer had the side-effect of not outputting prefixed types (like ui.Color) because they are no longer unknown, but the I've updated the type name finders so they always ask the AST as a workaround.

Fixes #130

### DIFF
--- a/mobx_codegen/test/data/valid_annotated_store_referencing_store_input.dart
+++ b/mobx_codegen/test/data/valid_annotated_store_referencing_store_input.dart
@@ -1,8 +1,8 @@
 library generator_sample;
 
-import 'dart:ui' as ui;
-
 import 'package:mobx/mobx.dart';
+
+import 'dart:ui';
 
 part 'generator_sample.g.dart';
 
@@ -11,7 +11,7 @@ class _Car {
   _Car(this.engine);
 
   @observable
-  ui.Color paintColor;
+  Color paintColor;
 
   @observable
   Engine _engine;
--- a/mobx_codegen/CHANGELOG.md
+++ b/mobx_codegen/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 0.3.9+1 - 0.3.10
+
+- Alters the analyzer dependency to support a range — from the previously supported version
+  (0.36.3), up to latest (0.39.0).
+- Adds support for library prefixes in all situations (`import 'package:foo' as foo`),
+  so type names are prefixed in generated part files.
+
 ## 0.3.8 - 0.3.9+1
 
 - Fixes a minor issue where types in generated code would appear as dynamic when they shouldn't.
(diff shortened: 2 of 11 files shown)


mobx_codegen/test/data/valid_annotated_store_referencing_store_input.dart
```diff
@@ -1,17 +1,17 @@
 library generator_sample;
 
-import 'dart:ui' as ui;
-
 import 'package:mobx/mobx.dart';
 
+import 'dart:ui';
```
pavanpodila: I think we also need  a test which includes the prefix and ensure that works too. Not sure if we can have a hybrid approach?

shyndman: Right, so it won't work anymore. In fact, prefixes only ever worked when the types were detected as dynamic, like with dart:ui types, but not in other situations.

You can see for yourself. If you open up `valid_input.dart` in the `mobx_codegen` tests, and import dart:core like this:

```dart
import 'dart:core' as c;

abstract class UserBase with Store {
  UserBase(this.id);

  final c.int id;

  @observable
  c.String firstName = 'Jane';

  @observable
  c.String lastName = 'Doe';

  @computed
  c.String get fullName => '$firstName $lastName';

...
```
You'll see that `valid_output.dart` doesn't require changing to get the test to pass, yet it contains unprefixed core types. In a real situation it wouldn't compile properly alongside the input.

I propose we fix prefixes across the board.

pavanpodila: When you say fix across the board, are you referring to removing the prefix everywhere?

shyndman: No, adding support for them.

I've pushed a new version that adds support, as well as a test.

shyndman:
OK, changelog and version bump in.

I've upgraded the type name finder so that it always gets information from the AST instead of the element tree, so library prefixes now work in all situations.

pavanpodila:
Published `mobx_codegen`: https://pub.dev/packages/mobx_codegen. Thanks @shyndman, this is a great step forward!

PR #311, opened 2019-10-28
Change a test in a way that should fail

shyndman:
## DO NOT SUBMIT. For demonstration only.

This change should cause the related test to fail, because `valid_output.dart` hasn't been updated to include the `c.` prefixes. But the test doesn't fail. Why not? 

It's because prefixes aren't properly supported by mobx.dart, **and never were**, except for in the case where I made them work for "unknown" types, in #273

This is what I'm talking about in the other PR.


shyndman:
I upgraded the type name finding code to rectify this situation. Closing.

Issue #316, opened 2019-11-19
How to set up a Reaction?


Magenda:
Hi, dart/mobx newbie here...

I've read almost everything I could find about reactions (mobx website guide, github examples, stackoverflow, mobx articles, github relevant issues etc) but I don't get how to properly set them up. 

For example, the following reaction just listens to an observable number and prints it's value when the latter gets modified. The following code has this syntax error for my number: "Only static members can be accessed in initializers.". What I'm doing wrong? 

```dart
import 'package:mobx/mobx.dart';
part 'live.g.dart';

class Live = _Live with _$Live;

abstract class _Live with Store {
  @observable
  int mynum = 0;

  // print mynum every time it's changed...
  final myreaction = reaction((_) => mynum, (msg) => print("$msg"));

  _Live() {
    mynum++; //just for testing the reaction...
  }

  @override
  dispose() {
    myreaction();
  }
}
```

pavanpodila:
One option is to setup these reactions in the constructor, eg:

```dart
abstract class _Live with Store {

  ReactionDisposer _disposer;
  
  _Live() {
    _disposer = reaction((_) => mynum, (msg) => print("$msg"));
  }

  @override
  dispose() {
    _disposer();
  }
}```

Magenda:
@pavanpodila I see, many thanks!

Is this the "official" way? I would love to know other common ways of usage. 

Moreover, the [GetStarted reactions section](https://mobx.pub/concepts#reactions) shows a usage outside of a class. Is this on purpose and a common way to import observables/actions/reactions from other files to the Store, or is it class-less just for code minimization in the guide?

MaikuB:
You can definitely do within a class e.g. a widget. I think the examples appear to be done outside of a class because it was only to show the relevant code. At least that's my assumption.

An example would be within a`StatefulWidget` for a page (i.e. one that has a scaffold), you can setup a reaction that listens to connectivity changes. When the user is offline show a message via snackbar. Override the dispose method so that you can dispose the reaction.

@pavanpodila I can try to update our examples to include this in, let me know what you think :)

pavanpodila:
That would be awesome @MaikuB. I think we should have a completely new example just to illustrate the various reactions: `autorun`, `reaction` and `when`. What do you say ? PR time ?

MaikuB:
[...]

PR #319, opened 2019-11-23
add connectivity example that demonstrates using a reaction

MaikuB:
added a new example as discussed in #316

### DIFF
--- a/mobx_examples/lib/connectivity/connectivity_store.dart
+++ b/mobx_examples/lib/connectivity/connectivity_store.dart
@@ -1,0 +1,15 @@
+import 'package:mobx/mobx.dart';
+import 'package:connectivity/connectivity.dart';
+
+part 'connectivity_store.g.dart';
+
+class ConnectivityStore = _ConnectivityStore with _$ConnectivityStore;
+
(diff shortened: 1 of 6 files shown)


mobx_examples/lib/connectivity/connectivity_store.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:mobx/mobx.dart';
```
pavanpodila: I think it would be good to have an example page, say `examples/connectivity` for [mobx.pub](https://mobx.pub) too. What do you think?

MaikuB: Can do though I suspect it'll be rather brief. Another issue is with the delay set to four seconds, I suspect not a lot of users will wait to see how it looks if we put a gif on there.

In any case I think that could be done in another PR. I did see that you have another PR for updating the site as well. Not sure if you plan to complete the soon or is that still in-progress?

pavanpodila: That PR was for upgrading the Docz tool. We can do the docs in a separate PR for the example

PR #320, opened 2019-11-24
add connectivity example to the docs

MaikuB:
following on from PR #319.

Not sure if other files are supposed to be modified but I was able to run the site locally (i.e. via `npm start`) and see the new page there


docs/src/examples/connectivity/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+}
+```
+
+The `ConnectivityStore` has a field named `connectivityStream`. It is an `ObservableStream` that wraps around a stream provided by the `connectivity` plugin that can be used to listen for changes in connection state.
+By doing so, we can make use of reactions to respond to changes as we shall soon see. Without MobX, this would generally have been done by [listening to the stream](https://api.dart.dev/stable/dart-async/Stream/listen.html).
+
+## Setting up the reaction
+
+To have our user interface inform users of changes about the state of their connection, we'll need to configure a reaction.
+We'll need to '`StatefulWidget` so that the reaction can be configured within the `initState` method and disposed of when the state itself is disposed. The complete code is as follows
```
pavanpodila: Minor grammatical errors: 

- "We'll need to use a StatefulWidget..." 
- disposed off (double f)

docs/src/examples/connectivity/index.mdx
```diff
@@ -0,0 +1,127 @@
+---
```
pavanpodila: Excellent article @MaikuB. Wonderfully written and easy to follow 🥇

docs/src/examples/connectivity/index.mdx
MaikuB: done. thanks for picking that up

docs/src/examples/connectivity/index.mdx
MaikuB: cheers :)

docs/src/examples/connectivity/index.mdx
MaikuB: Oh noticed you wrote "disposed off". I haven't seen anyone use that in this context here in Australia. A search online showed one result where someone was taught that in India. See https://forum.wordreference.com/threads/dispose-of-dispose-off.556486/

MaikuB: I've reworded it so this shouldn't be as relevant anymore :)

docs/src/examples/connectivity/index.mdx
```diff
@@ -46,7 +46,7 @@ By doing so, we can make use of reactions to respond to changes as we shall soon
 ## Setting up the reaction
 
 To have our user interface inform users of changes about the state of their connection, we'll need to configure a reaction.
-We'll need to '`StatefulWidget` so that the reaction can be configured within the `initState` method and disposed of when the state itself is disposed. The complete code is as follows
+We'll need to use a `StatefulWidget` so that the reaction can be configured within the `initState` method and disposed of when the state itself is going to be disposed. The complete code is as follows
```
pavanpodila: Can reword "disposed of" to just "disposed"

MaikuB: This comment is against an outdated version. Have you looked at the most recent version? It's broken up into two sentences now. Or did you see that and now suggesting I go back to what it was before and simply remove "of"?

If I'm not mistaken, my PR should be open for edits that you could make the edits as well. Feel free to do so if you wish :)

pavanpodila: I think the sentence is good for now :-). Thanks for writing up the example and doc so promptly!

PR #334, opened 2019-12-18
Fix: Widgets failed to call super.dispose()

rchavik:
When using a StatefulWidget, I would like to use its State object as a
Store:

```dart

part 'test_widget.g.dart';

class TestWidget extends StatefulWidget {
  @override
  _TestWidgetState createState() => _TestWidgetState();
}

class _TestWidgetState = _TestWidgetStore with _$_TestWidgetState;

abstract class _TestWidgetStore extends State<TestWidget> with Store {

  @observable
  int counter;

  @override
  Widget build(BuildContext context) {
    return Container();
  }
}
```

This already works but during runtime, there are error message such as: `Widget.foo failed to call super.dispose()`.

Removing dispose() from the Store mixin seems like a good solution.
Caller can still override dispose on their parent class.

Possibly related:
https://github.com/flutter/flutter/issues/24293
https://github.com/dart-lang/sdk/issues/27100


shyndman:
@pavanpodila 

This is technically a breaking change...but I wonder if it is in practice. Does anyone use `dispose()`?

I would love to get rid of `dispose()`. It isn't useful in its current form. Maybe we could go 1.0? Consider it breaking?

shyndman:
@rchavik — It's unusual to be using a `State` as a `Store`. Have you considered externalizing your "Store" state into an external class? Have you looked over [the examples](https://mobx.pub/examples/counter)?

rchavik:
> This is technically a breaking change...but I wonder if it is in practice. Does anyone use dispose()?

How is it breaking change? ie: anyone who override dispose() in their store should still work.

> @rchavik It's unusual to be using a State as a Store. Have you considered externalizing your "Store" state into an external class?

Yes, I do use external Store classes for shared states.  What are widget state classes aside acting as a internal widget Store? I think it makes sense in some cases.

rchavik:
> How is it breaking change? ie: anyone who override dispose() in their store should still work.

Ah. In some cases, they'd have to manually remove the @override decorator.

shyndman:
> How is it breaking change? ie: anyone who override dispose() in their store should still work.

Couple ways. It could be that someone is doing something that amounts to `(foo as Store).dispose()`. Another way this might manifest is in a Store subclass, `super.dispose()`. It's a public interface, despite it being unrelated to the functioning of the library.

> Yes, I do use external Store classes for shared states. What are widget state classes aside acting as a internal widget Store? I think it makes sense in some cases.

I don't doubt it. I'm just curious. Can you describe how you use it a little more? Like a toy example?

rchavik:
> I don't doubt it. I'm just curious. Can you describe how you use it a little more? Like a toy example?

Eg: i have a PostsStore in `stores/posts_store.dart`:
```dart
class PostsStore = _PostsStore with _$PostsStore;
abstract class _PostsStore with Store {
   @observable
   List<Post> cachedPosts;
   @action
   getPosts() async {
      cachedPosts = api.getPosts();
   }
   getRelated(Post post) async {
      return api.getRelated(post);
   }
}
```

PostsStore.cachedPosts is shared between several widgets.  All is great. 

Now i have to display a list of related posts given a Post instance.  It doesn't really makes sense to store this in the external store as it depend on a post.  Sure, I could use a ObservableMap and store a list of related posts based on a post.id as key.

But then the shared external store can be bloated for this 'small' use case.  We can create a widget exclusive store, but then it adds an extra file for the store and another for the generated code.

So, I think why not simply use the widget's State object as a Store? ie.:

```dart

class RelatedPosts extends StatefulWidget {
  final int id;
  RelatedPosts({Key key, this.id}) : super(key: key);
  _RelatedPostsState createState() => _RelatedPostsState();
}

abstract class _RelatedPostsStore extends State<RelatedPosts> with Store {
  @observable relatedPosts = [];
  @override
  void didChangeDependencies() {
    super.didChangeDependencies();
    final postStore = Provider.of<PostStore>(context);
    postStore.getRelated(widget.id).then((results) => relatedPosts = results);
  }
  Widget build() {
     return ListView.builder(
       items: relatedPosts,
     );
  }
}

**Comment by shyndman** [comment]
In order for MobX to know when to re-render the widget, it needs to track dependencies, and for that it needs a reaction (`Observer` in Flutter-land), which I don't see here.

Does this work as is? Or is there an `Observer` used in the `build` method?

**Comment by rchavik** [comment]
> In order for MobX to know when to re-render the widget, it needs to track dependencies, and for that it needs a reaction (`Observer` in Flutter-land), which I don't see here.
> 
> Does this work as is? Or is there an `Observer` used in the `build` method?

You're right, I missed that.  Have now updated the example code.

**Comment by shyndman** [comment]
Thank you. 

First off, I appreciate the perspective on this. It's an interesting approach.

This PR, however, requires a bit of thought from a library perspective. I'm going to let @pavanpodila pipe in on whether and how we can accept the change.

In the interim, here's an approach I'd consider (your call entirely, but it might be more understandable given the context).

Since you're already dealing with a `State` object, a re-render can be triggered with a call to `setState()` in place of the `@observable`+`Observer` combo. It accomplishes the same with way less machinery. It has the advantage of being Flutter's standard way of re-rendering after a change in a `StatefulWidget`. It might look something like this:

```dart
class RelatedPosts extends StatefulWidget {
  final int id;
  RelatedPosts({Key key, this.id}) : super(key: key);
  _RelatedPostsState createState() => _RelatedPostsState();
}

class _RelatedPostsStore extends State<RelatedPosts> {
  var relatedPosts = <Post>[];

  @override
  void didChangeDependencies() {
    super.didChangeDependencies();
    final postStore = Provider.of<PostStore>(context);
    postStore.getRelated(widget.id).then((results) {
      //!
      setState(() => relatedPosts = results);
    });
  }

  @override
  Widget build(BuildContext context) {
    return ListView.builder(
      items: relatedPosts,
    );
  }
}
```

shyndman:
Oh, and I have a pure MobX solution that you might like. Give me a sec to write it up.

shyndman:
> Now i have to display a list of related posts given a Post instance. It doesn't really makes sense to store this in the external store as it depend on a post. Sure, I could use a ObservableMap and store a list of related posts based on a post.id as key.

Nevermind, I get it. And I see you know what you're doing. 

If this PR doesn't go through, I'd stick with my suggestion, or have an external Store class. In the same file probably, if you were concerned about directory bloat.

rchavik:
> First off, I appreciate the perspective on this. It's an interesting approach.

I was trying to duplicate this for flutter: https://medium.com/react-native-training/ditching-setstate-for-mobx-766c165e4578

It uses @observable within the Component class itself, and does not @inject stores to the component.

shyndman:
Huh. Interesting. That does seem like a nearly identical use-case. 

Still feels strange to me. I wonder if the JS MobX team would recommend a similar approach.

shyndman:
Regardless! Let's get rid of dispose. I think you should be able to do this.

rrousselGit:
You can get rid of dispose. It should be a problem

On the other hand, your way of making StatefulWidget work as an observer in incorrect.


There are better solutions. I'll make a separate PR.

pavanpodila:
I agree with the idea of getting rid of `dispose()`, I don't think the Store mixin should include it. For Stores that do need the lifecyle, they can add their own and call its dispose when needed. Also agree with @rrousselGit that we should avoid using State directly inside the Widget. 

Externalizing the Store gives you benefits in terms of unit testing and being independent of  the UI.

shyndman:
In that case, let's review this PR. And figure out how we want to version afterwards.

mobx/test/store_test.dart
```diff
@@ -6,8 +6,5 @@ class TestStore with Store {}
 void main() {
```
shyndman: One thing I might suggest is getting rid of this file entirely. This was the only test, right? We can add the file back when required.

pavanpodila: Yes agree, let's delete this test file

rchavik:
@shyndman @pavanpodila I have removed the file as requested. Thanks for bearing with me.

@rrousselGit I'm still interested in your alternative way about making StatefulWidget works as an observer.

PR #336, opened 2019-12-18
Added StatelessObserverWidget and StatefulObserverWidget

rrousselGit:
This adds two new kinds of widgets:
- StatelessObserverWidget
- StatefulObserverWidget

And refactor `Observer` to simply subclass `StatelessObserverWidget`.

Basically, these new widgets are simply `StatelessWidget`/`StatefulWidget` where their `build` method tracks observables.

This works by subclassing their respective `Element`.

It also removes the const constructor on `Observer`, which cannot be used anyway.

It also refactors `createReaction` to create the Reaction's name based on `widget.toString` instead of `context.nameFor`.

This gives better names (`class MyWidget extends StatelessObserverWidget` will name the reaction `MyWidget#someHashCode`). 
And it removes the weird behavior where `ReactiveContext.nextId` kept increasing for no reason.

And it also slightly modify `analysis_options.yaml` to disable some conflicting rules, and suggest to update `mobx` to disable both `implicit-dynamic` and `implicit-cast`.


rrousselGit:
cc @rchavik

rrousselGit:
Basically, before:

```dart
class MyWidget extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return Observer(
      builder: (context) {
        ...
      },
    );
  }
}
```

After:

```dart
class MyWidget extends StatelessObserverWidget {
  @override
  Widget build(BuildContext context) {
    ...
  }
}
```

pavanpodila:
The `Observer` is also a required widget since you may **not** want to subclass all the time. In fact, it eliminates the need to subclass in most cases and you can inject `Observer` just at the places you need. With this proposed solution, you are forcing the users to subclass ?

Can I still use the `Observer` with the `builder` ?

rrousselGit:
`Observer` is still there. It's just that now its implementation is very simple:

```dart
class Observer extends StatelessObserverWidget {
  Observer({Key key, this.builder}): super(key: key);

  final WidgetBuilder builder;

  @override
  Widget build(BuildContext context) => builder(context);
}
```

pavanpodila:
nm...got my answer. We can still use the Observer. 

My question is: "When would the StatefulObserver be useful ?"

analysis_options.yaml
```diff
@@ -1,12 +1,23 @@
+analyzer:
+  exclude:
+    - mobx_examples/lib/**/*.g.dart
+    - mobx_codegen/test/data/**.dart
+  # TODO: migrate mobx to enable these
+  # strong-mode:
+  #   implicit-casts: false
```
pavanpodila: What is preventing from using this right now ?

rrousselGit:
The stateful variant is useful for things like animation, or if it uses a
local Store

pavanpodila:
Ah...animations... This is a place where MobX can really shine, as in whenever an observable property changes, we can trigger automatic animations, all via plain reactions. I haven't explored this use case yet but seen it being used in the JS world. Need to see how well we can create an API for MobX based animations. 

We could do it the way SwiftUI does it using `withAnimation { ... }`. We can use it internally to track changes and trigger animations!

analysis_options.yaml
rrousselGit: Enabling both triggers more than 100 errors

They should be fixed, but probably not as part of this PR

rrousselGit:
> Looks good overall. There is some drop in coverage. Can we fix that and bring it back to 98%?

Ah, it was just a typo
Fixed

rrousselGit:
> Ah...animations... This is a place where MobX can really shine, as in whenever an observable property changes, we can trigger automatic animations, all via plain reactions. I haven't explored this use case yet but seen it being used in the JS world. Need to see how well we can create an API for MobX based animations.
>
> We could do it the way SwiftUI does it using withAnimation { ... }. We can use it internally to track changes and trigger animations!

I don't think that's the job of `mobx` in the Flutter world.

Flutter comes with an implicit animation framework already. Mobx is just an `Observable` on the top of these implicitly animated widgets.

rchavik:
@rrousselGit 

Am I correct in on how to use the new `StatefulObserverWidget`?

```dart
part 'related_post.g.dart';

class RelatedPosts extends StatefulObserverWidget {
  final int id;
  RelatedPosts({Key key, this.id}) : super(key: key);
  _RelatedPostsState createState() => _RelatedPostsState();
}

class _RelatedPostsState = _RelatedPostsStore with _$_RelatedPostsState;

abstract class _RelatedPostsStore extends State<RelatedPosts> with Store {
  @observable relatedPosts = [];
  @override
  void didChangeDependencies() {
    super.didChangeDependencies();
    final postStore = Provider.of<PostStore>(context);
    postStore.getRelated(widget.id).then((results) => relatedPosts = results);
  }
  Widget build() {
     return ListView.builder(
       items: relatedPosts,
     );
  }
}
```

The main difference here is that we do not need to wrap the result of State.build with an Observer().

Braindump: Is it possible to make mobx_codegen support the following:

```dart
part 'related_post.g.dart';

class RelatedPosts extends StatefulObserverWidget {
  final int id;
  RelatedPosts({Key key, this.id}) : super(key: key);
  _RelatedPostsState createState() => _RelatedPostsState();
}

class _RelatedPostsState extends ObservableState<RelatedPosts>  {
  @observable relatedPosts = [];
  @override
  void didChangeDependencies() {
    super.didChangeDependencies();
    final postStore = Provider.of<PostStore>(context);
    postStore.getRelated(widget.id).then((results) => relatedPosts = results);
  }
  Widget build() {
     return ListView.builder(
       items: relatedPosts,
     );
  }
}
```

When a mobx_codegen sees an `ObservableState`, it generates the mixin much like it generates the part file when it sees the `Store` mixin.   If so,  we can do away with the class aliasing `class _RelatedPostsState = _RelatedPostsStore with _$_RelatedPostsState;`

pavanpodila:
@rrousselGit I think it would be good to add some docs around these new `Observer` classes. Would be easier for users to get started using it.

PR #342, opened 2019-12-24
1-added (ObserverProvider) & (ObserverStatefulProvider) widgets, they…

kaarimtareek:
… are like the default Observer but with easy way for binding the ui with the view models(stores) , this pattern is inspired from Filled Stacks package view model provider.2- added mobxbase class which  you should inheret from when creating your stores viewmodels.

### DIFF
--- a/flutter_mobx/lib/flutter_mobx.dart
+++ b/flutter_mobx/lib/flutter_mobx.dart
@@ -4,5 +4,7 @@ library flutter_mobx;
 
 export 'package:flutter_mobx/src/observer.dart'
     hide debugAddStrackTraceInObserverName;
+export 'package:flutter_mobx/src/mobx_base.dart';
+export 'package:flutter_mobx/src/observer_provider.dart';
 
 const version = '0.3.4+1';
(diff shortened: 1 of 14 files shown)


pavanpodila:
@kaarimtareek Please check the conflicts and also the big drop in code coverage. I think you will need to add tests as well.

rrousselGit:
I don't think that's a good thing to add these directly v to b mobx

These probably make more sense as a package, especially with the code being that simple

pavanpodila:
Agree with @rrousselGit. You should publish this as a separate package, say `mobx_provider` or something along those lines. BTW, @kaarimtareek, you should know that @rrousselGit is the maintainer of the Provider package and has worked on a similar solution in the past.

I am going to close this PR in favor of an external package.

kaarimtareek:
thank you for your feedbacks , I have two questions ,
will the package be part of mobx.dart package or will be a complete
different package?
the second one is about naming the mobx classes , should i name them with
'store' or 'viewmodel' ? just to be consistent with your pattern and not
creating a confusion to the users?

PR #344, opened 2019-12-26
Print full error details in test scenarios

### LINKED ISSUE #343: Improve code generator error messages
When the generator fails, it spits out an unhelpful "Error in StoreGenerator". It would improve the debugging and bug reporting situation if we printed something more meaningful.

shyndman:
Previously the generator would fail with 'Error running StoreGenerator'
in test situations, adding overhead to figuring out the source of
errors.

Closes #343


shyndman:
Thanks!

Issue #345, opened 2019-12-26
Create ObservableStream Extension


jacobaraujo7:
It would be really cool if there was an extension to the ObservableStream Stream like:

```dart
ObservableStream output = stream.asObservable();
```
instead

```dart
ObservableStream output = ObservableStream(stream);
```

pavanpodila:
@jacobaraujo7 Want to try a PR? 👍

jacobaraujo7:
@pavanpodila 
Yes sir! What do you think about this? Is it also worth thinking about extensions for other collections like ObservableList?

pavanpodila:
Yes, extensions is a huge opportunity to really simplify the API and make it more intuitive. I would certainly explore other APIs that can be introduced, for Observable, Computed, Atom, Observable{List,Set,Map}, ObservableFuture and possibly even inside the core

jacobaraujo7:
I believe the ObservableStream extension will bring BLoC and MobX into the community.

jacobaraujo7:
@pavanpodila If you get your endorsement to work with the minimum dart version 2.7.0, then I will start working on the other extensions.

pavanpodila:
Closing it as part of your PR #346

PR #346, opened 2019-12-26
added ObservableStream, ObservableFuture and Collections extension #345

jacobaraujo7:
* [x] Create extension for Stream added feature "asObservable()".
* [x] Create extension for Future added feature "asObservable()".
* [x] Create extension for List added feature "asObservable()".
* [x] Create extension for Set added feature "asObservable()".
* [x] Create extension for Map added feature "asObservable()".

```dart
ObservableStream output = stream.asObservable();
```
instead

```dart
ObservableStream output = ObservableStream(stream);
```

### DIFF
--- a/mobx/lib/mobx.dart
+++ b/mobx/lib/mobx.dart
@@ -61,6 +61,7 @@ export 'package:mobx/src/core.dart'
         NotificationHandlers,
         ReactionImpl,
         DerivationState;
+export 'package:mobx/src/api/extensions.dart';
 
 /// The current version as per `pubspec.yaml`
 const version = '0.3.10';
--- a/mobx/lib/mobx.dart
+++ b/mobx/lib/mobx.dart
@@ -64,4 +64,4 @@ export 'package:mobx/src/core.dart'
 export 'package:mobx/src/api/extensions.dart';
 
 /// The current version as per `pubspec.yaml`
-const version = '0.3.10';
+const version = '0.4.0';
--- a/mobx/lib/src/api/extensions/observable_map_extension.dart
+++ b/mobx/lib/src/api/extensions/observable_map_extension.dart
@@ -2,7 +2,5 @@ part of '../extensions.dart';
 
 /// Turn the Map into an ObservableMap.
 extension ObservableMapExtension<K, V> on Map<K, V> {
-  ObservableMap<K, V> asObservable() {
-    return ObservableMap<K, V>.of(this);
-  }
+  ObservableMap<K, V> asObservable() => ObservableMap<K, V>.of(this);
 }
--- a/mobx/lib/src/api/extensions/observable_map_extension.dart
+++ b/mobx/lib/src/api/extensions/observable_map_extension.dart
@@ -1,0 +1,8 @@
+part of '../extensions.dart';
+
+/// Turn the Map into an ObservableMap.
+extension ObservableMapExtension<K, V> on Map<K, V> {
+  ObservableMap<K, V> asObservable() {
+    return ObservableMap<K, V>.of(this);
+  }
+}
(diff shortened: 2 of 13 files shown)


shyndman:
What implication does this have to our minimum supported version?

jacobaraujo7:
to ensure feature usage, but not required for use. do you want me to return the minimum version of dart to 2.2.0?

mobx/lib/mobx.dart
```diff
@@ -61,6 +61,7 @@ export 'package:mobx/src/core.dart'
         NotificationHandlers,
         ReactionImpl,
         DerivationState;
+export 'package:mobx/src/api/extensions.dart';
 
 /// The current version as per `pubspec.yaml`
 const version = '0.3.10';
```
pavanpodila: We will need to bump the version in `pubspec.yaml`

pavanpodila:
I think for this feature to go out, the minimum version will have to be **2.7**

jacobaraujo7:
My bad!

jacobaraujo7:
I researched a lot about it, and couldn't find a way to import conditionally based on the SDK version.
Apparently the only way is to change the minimum version to 2.7, following the steps of other packages like rxdart.

jacobaraujo7:
I moved and added extension to ObservableFuture and Collections (list, map, set)

mobx/pubspec.yaml
```diff
@@ -9,7 +9,7 @@ authors:
 homepage: https://github.com/mobxjs/mobx.dart
 
 environment:
-  sdk: '>=2.2.0 <3.0.0'
+  sdk: '>=2.7.0 <3.0.0'
```
pavanpodila: Can you educate me on how rxdart has done this? We are forcing everyone to move up to dart 2.7, which is fine since extensions can't be used otherwise. Curious how other packages are dealing with this?

mobx/lib/src/api/extensions/observable_map_extension.dart
```diff
@@ -0,0 +1,6 @@
+part of '../extensions.dart';
+
+/// Turn the Map into an ObservableMap.
+extension ObservableMapExtension<K, V> on Map<K, V> {
+  ObservableMap<K, V> asObservable() => ObservableMap<K, V>.of(this);
```
shyndman: Let's include named, optional context parameters for all `asObservable` methods.

shyndman: Like how you're doing it with `Future`s and `Stream`s.

mobx/pubspec.yaml
jacobaraujo7: In the case of rxdart, they proposed using extensions to replace the internal wrappers when the feature was in BETA. When Flutter 1.12 with dart 2.7 support was released, they updated the package generating Breaking Changes, but it was necessary to use these new features as extensions and raise the lib threshold.

I believe that with MobX the story will be different as it will be idol for Flutter as developers are always very close to the stable version and always update.

I have doubts as to other platforms that use Dart, in this environment I do not know how this is applied.

mobx/lib/src/api/extensions/observable_map_extension.dart
jacobaraujo7: Added!

mobx/pubspec.yaml
pavanpodila: I am happy to upgrade if folks are always on latest Flutter. @shyndman, @rrousselGit any insights I should know before I bump the dart version to **2.7**?

shyndman: I think it should be fine. People can always reference an earlier version if needed.

pavanpodila: Great, let's do this with min as 2.7

mobx/lib/mobx.dart
pavanpodila: Let's make the version as 0.4.0, as we are bumping up the Dart SDK.

jacobaraujo7: done!

jacobaraujo7:
@pavanpodila tests done!

mobx/lib/mobx.dart
jacobaraujo7: done

pavanpodila:
Cool stuff @jacobaraujo7 🏆

PR #350, opened 2019-12-26
Add generated test coverage badge to gitignore

shyndman:
This file is generated as a result of a `pub run test_coverage`, and is unneeded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ Generated.xcconfig
 
 coverage/
 .test_coverage.dart
+coverage_badge.svg
 
 # Miscellaneous
 *.class


.gitignore
```diff
@@ -31,6 +31,7 @@ Generated.xcconfig
 
 coverage/
 .test_coverage.dart
+coverage_badge.svg
```
pavanpodila: How is this relevant to us ?

shyndman: If you run test_coverage locally, it gets generated. 🤷‍♂️

pavanpodila: Umm....ok

PR #365, opened 2019-12-30
Correct Observer debug name to point to constructing frame

### LINKED ISSUE #361: Observer is no longer named according to its location in user source
My fault for not having a test in place.

shyndman:
NOTE: This merged into the wrong branch last time around.

Having been moved into a getName() call, the function failed to serve
its purpose of pointing to a location in user source for easier
debugging.

Determination of the stack frame has also been improved to search for
the first non-constructor, in order to support subclasses of Observable.
This functionality has been tested in the browser and VM.

Fixes #361

MR: constructing_frame2

### DIFF
--- a/flutter_mobx/lib/src/observer.dart
+++ b/flutter_mobx/lib/src/observer.dart
@@ -1,7 +1,16 @@
 // ignore_for_file:implementation_imports
+import 'dart:convert';
+
 import 'package:flutter/widgets.dart';
 import 'package:mobx/mobx.dart';
 import 'package:mobx/src/core.dart' show ReactionImpl;
+
+/// `true` if a stack frame indicating where an [Observer] was created should be
+/// included in its name. This is useful during debugging to identify the source
+/// of warnings or errors.
+///
+/// Note that stack frames are only included in debug builds.
+bool debugAddStackTraceInObserverName = true;
 
 /// A [StatelessObserverWidget] that delegate its [build] method to [builder].
 ///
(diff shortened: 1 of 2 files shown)


flutter_mobx/lib/src/observer.dart
```diff
@@ -13,15 +22,54 @@ class Observer extends StatelessObserverWidget
...
+  static String debugFindConstructingStackFrame([StackTrace stackTrace]) {
+    String stackFrame;
+
+    assert(() {
+      if (debugAddStackTraceInObserverName) {
+        final stackTraceString = (stackTrace ?? StackTrace.current).toString();
+        stackFrame = LineSplitter.split(stackTraceString)
+            // Evalulate frames starting from below this closure, method, and
+            // Observer constructor
+            .skip(3)
```
pavanpodila: What is the logic for skipping three levels. I think we should document it with some comments.

pavanpodila: Also how will this work if someone decides to use the `StatelessObserverWidget` or `StatefulObserverWidget` ?

shyndman: The comment describing the skip is right above the line. How should I clarify?

shyndman: I made a small change to the comment. That make more sense?

shyndman: Regarding the observer widgets, it doesn't — the `const` constructors prevent it.

shyndman: That, or we introduce a non-const constructor to each of those classes, which will populate a stack frame variable. Your call.

pavanpodila: > That, or we introduce a non-const constructor to each of those classes, which will populate a stack frame variable. Your call.

Ya we should ensure a user of the Subclass gets the right location as well. I think we should add the constructors to them.

pavanpodila: > The comment describing the skip is right above the line. How should I clarify?

I saw that. What I was curious was about the magic number **3**. Why is that relevant and can we qualify its use here ? As in why are we skipping 3 levels and not 4. Is there a hidden assumption here?

shyndman: 1. the anonymous function in the assert
2. the `debugFindConstructingStackFrame` method
3. the constructor invoking that `debugFindConstructingStackFrame`
4. either user source (which is what we want), or an Observer subclass' constructor (which we skip past with the regex)

shyndman: > > That, or we introduce a non-const constructor to each of those classes, which will populate a stack frame variable. Your call.
> 
> Ya we should ensure a user of the Subclass gets the right location as well. I think we should add the constructors to them.

ok. i'll take a look at hoisting it tomorrow.

pavanpodila: > 1. the anonymous function in the assert
> 2. the `debugFindConstructingStackFrame` method
> 3. the constructor invoking `debugFindConstructingStackFrame`
> 4. either user source (which is what we want), or an Observer subclass' constructor (which we skip past with the regex)
> 
> Maybe I should include this rationale in the comment? To be very clear?

Yup! Right now we have lots of mental context, so we find this discovery trivial, but in few weeks, I am sure we will be scratching our heads again :-) as to why we skipped this many levels...

shyndman: Well, I mean, the comment already says that. Maybe just not clear enough. I'll amend.

shyndman: OK. Done.

PR #368, opened 2020-01-02
fix for the return-type when it is generic with type-arguments

pavanpodila:


mobx_codegen/lib/src/type_names.dart
```diff
@@ -75,20 +73,12 @@ class LibraryScopedNameFinder {
...
-  /// Calculates a type name, including its type arguments if
-  /// [includeTypeArguments] is `true`.
+  /// Calculates a type name, including its type arguments
   ///
   /// The returned string will include import prefixes on all applicable types.
-  String _getTypeName(DartType type, {bool includeTypeArguments = false}) {
+  String _getTypeName(DartType type) {
     final sb = StringBuffer()..write(_findUnparameterizedTypeName(type));
 
-    if (type is ParameterizedType && includeTypeArguments) {
```
pavanpodila: This was causing issues as it would loop through type arguments and generate an invalid type: such as `List<User><User>`

shyndman:
Hey! Thanks for looking at this. I was in the air all day and offline.

LGTM.

pavanpodila:
published on pub. And welcome back to land :-)

Issue #354, opened 2019-12-28
Upgrade analyzer dependency lower bound and remove deprecated API usage


shyndman:
`analyzer` 0.39.x introduced deprecations for APIs that we depend on. As we increase the dependency's lower bound, we can move to using the replacements.

**analyzer 0.38.2**
* Introduces `NullabilitySuffix`, required for calling `ClassElement.instantiate`, which can be used in place of `TypeDefiningElement.type`

**analyzer 0.39.1**
* Introduces `LibraryElement.typeSystem`, the replacement for `AnalysisSession.typeSystem`

**analyzer 0.39.2**
* Introduces `DartType.getDisplayString()`, the replacement for `DartType.displayName`

The real question here is how long we want to keep supporting older `analyzer` versions.

leonardocustodio:
Hi, is it possible to upgrade for analyzer 0.40.x ? Other libraries are already using (for example hive_generator) and both are conflicting.

pavanpodila:
Definitely possible. Do you want to do a PR? 🤞

leonardocustodio:
Ok. I can check it out, just give me a few minutes to check another thing first.

shyndman:
Nice work @leonardocustodio

PR #380, opened 2020-01-07
Ignore another analyzer deprecation

shyndman:
The analyzer 0.39.3 update was causing our latest builds to break, because of a new deprecation to `DartType.displayName`.

I updated #354 as a reminder to remove it when we upgrade our dependencies.

### DIFF
--- a/mobx_codegen/lib/src/type_names.dart
+++ b/mobx_codegen/lib/src/type_names.dart
@@ -96,6 +96,10 @@ class LibraryScopedNameFinder {
             type is TypeParameterType ||
             // If the type lives in this library, no prefix necessary
             typeElement.library == library) {
+      // TODO(shyndman): This ignored deprecation can be removed when we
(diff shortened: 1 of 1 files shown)


mobx_codegen/lib/src/type_names.dart
```diff
@@ -96,6 +96,10 @@ class LibraryScopedNameFinder {
             type is TypeParameterType ||
             // If the type lives in this library, no prefix necessary
             typeElement.library == library) {
+      // TODO(shyndman): This ignored deprecation can be removed when we
```
pavanpodila: What's preventing from increasing the min version ?

shyndman:
Let's avoid another minor version bump. We should work on stabilizing a version for people.

mobx_codegen/lib/src/type_names.dart
shyndman: Let's avoid another minor version bump. We should work on stabilizing a version for people.

PR #389, opened 2020-01-10
Dice example

giri-jeedigunta:
@pavanpodila I have added an example for newbies to help them understand the ease of using Mobx in flutter with a fun dice app example. Please review and let me know your thoughts and comments. Thanks!


pavanpodila:
Hi Giri,
       Can you also include the code in the `mobx_examples` project. We can link to it from the markdown writeup.

PR #391, opened 2020-01-11
mobx_codegen: allow generating toString method for Store class #174

hawkbee1:
There's still a lot to do.

Warning: I changed mobx_codegen/pubspec.yaml for testing purpose 
Warning: I put back a new version of _generateCodeForAnnotatedStore and I don't know why it has been removed in first place.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,7 @@ name: Publish
 
 on:
   push:
-    branches: 
+    branches:
       - master
 
 jobs:
@@ -26,7 +26,7 @@ jobs:
           suppressBuildRunner: true
 
       - name: Tag commit
-        if: steps.publish.outputs.remoteVersion != steps.publish.outputs.localVersion
+        if: steps.publish.outputs.success
         uses: hole19/git-tag-action@master
         env:
           TAG: ${{steps.publish.outputs.package}}-${{steps.publish.outputs.localVersion}}
@@ -53,7 +53,7 @@ jobs:
           suppressBuildRunner: true
 
       - name: Tag commit
-        if: steps.publish.outputs.remoteVersion != steps.publish.outputs.localVersion
+        if: steps.publish.outputs.success
         uses: hole19/git-tag-action@master
         env:
           TAG: ${{steps.publish.outputs.package}}-${{steps.publish.outputs.localVersion}}
--- a/.idea/workspace.xml
+++ b/.idea/workspace.xml
@@ -7,9 +7,8 @@
     </enabled-run-configurations>
   </component>
   <component name="ChangeListManager">
-    <list default="true" id="87dccf31-231e-4b6a-9b49-5967bcdc0ac7" name="Default Changelist" comment="added autorun, reaction example in cheat sheet">
+    <list default="true" id="87dccf31-231e-4b6a-9b49-5967bcdc0ac7" name="Default Changelist" comment="added when example in cheat sheet">
       <change beforePath="$PROJECT_DIR$/.idea/workspace.xml" beforeDir="false" afterPath="$PROJECT_DIR$/.idea/workspace.xml" afterDir="false" />
-      <change beforePath="$PROJECT_DIR$/docs/src/guides/cheat-sheet.mdx" beforeDir="false" afterPath="$PROJECT_DIR$/docs/src/guides/cheat-sheet.mdx" afterDir="false" />
     </list>
     <option name="SHOW_DIALOG" value="false" />
     <option name="HIGHLIGHT_CONFLICTS" value="true" />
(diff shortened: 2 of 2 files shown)


hawkbee1:
I don't understand the failed tests on CircleCI

shyndman:
> I don't understand the failed tests on CircleCI

They aren't you. There's something flaky in there. I triggered a re-run.

shyndman:
I wasn't involved in the original feature request, but mind if I drop a couple comments in?

mobx_codegen/lib/src/store_class_visitor.dart
```diff
@@ -90,6 +92,9 @@ class StoreClassVisitor extends SimpleElementVisitor {
       ..name = element.name;
 
     _storeTemplate.observables.add(template);
+    if(_storeTemplate.generateToString) {
+      _storeTemplate.toStringList.add('${element.name}: \${${element.name}.toString()}');
```
pavanpodila: I would prefer keeping track of the observable field names, which we already have with the` _storeTemplate.observables`. I think we can just use that to generate the `toString()` value.

mobx_codegen/test/data/valid_input_to_string_true.dart
```diff
@@ -0,0 +1,84 @@
...
+import 'package:mobx/mobx.dart';
+//import 'package:flutter/material.dart';
+
+part 'generator_sample.g.dart';
+
+typedef VoidCallback = void Function();
+
+class User = UserBase with _$User;
+
+@MakeStore(generateToString: true)
```
pavanpodila: We don't have store annotations anymore

mobx_codegen/lib/src/store_class_visitor.dart
hawkbee1: Right that's better. I do the same with computed ?

hawkbee1:
I will change all of this. Since we won't have store annotation anymore should I create a @toString annotation ?

mobx_codegen/lib/src/store_class_visitor.dart
pavanpodila: ya we can include computed as well

mobx_codegen/test/data/valid_input_to_string_true.dart
pavanpodila: > I will change all of this. Since we won't have store annotation anymore should I create a @tostring annotation ?

I think its still worth having a toString. Users can always override if they want to specialize differently.

shyndman: Except they'd have to override in the public class, which most people write without a body.

As a user, I'd prefer to have this flagged. Having it on by default will add a lot of work for me.

mobx_codegen/lib/src/template/store.dart
```diff
@@ -56,6 +60,30 @@ abstract class StoreTemplate {
...
+    if (generateToString) {
+      for(var i = 0; i < observables.templates.length; i++){
+        final ObservableTemplate current =  observables.templates[i];
+        toStringList.add('${current.name}: \${${current.name}.toString()}');
+      }
+
+      for(var i = 0; i < computeds.templates.length; i++){
+        final ObservableTemplate current =  computeds.templates[i];
+        toStringList.add('${current.name}: \${${current.name}.toString()}');
+      }
```
hawkbee1: I think about creating a Template class with name. ObservableTemplate and ComputedTemplate would extend this class and I could do something like
```dart
for(var i = 0; i < T.templates.length; i++){
        final T current =  computeds.templates[i];
        toStringList.add('${current.name}: \${${current.name}.toString()}');
      }
```

mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -33,6 +33,8 @@ class StoreGenerator extends Generator {
     for (final classElement in library.classes) {
       if (isMixinStoreClass(classElement)) {
         yield* _generateCodeForMixinStore(library, classElement, typeSystem);
+      } else if (isAnnotatedStoreClass(classElement)) {
```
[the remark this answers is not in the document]
hawkbee1: removed

mobx_codegen/test/data/valid_input_to_string_true.dart
pavanpodila: The other option is we introduce a `@codegen({toString: true})` style of annotation, which can be extended in the future ?

mobx_codegen/lib/src/template/store.dart
pavanpodila: I don't think you should extend `ObservableTemplate` or `ComputedTemplate`. Instead add, a **`toFieldValueString`** as a getter and use that as part of the `toString` inside the `StoreTemplate`

PR #398, opened 2020-01-16
Fix naming of tracking function argument in reaction

MaikuB:
Noticed that the argument in `reaction` was named as `predicate` when it doesn't return a boolean value. Changed this to `fn` though we may want to change this something else. Alternatives I can think of are `trackingFunction`/`trackingFn` or `dataFunction`/`dataFn`

Sidenote: suggest agreeing on a standard in the API docs. I noticed there are cases where functions are referred to as `<function>` and other places use `<function>()`. I would suggest the former and it's also the approach used in the Flutter docs

### DIFF
--- a/docs/src/api/reaction.mdx
+++ b/docs/src/api/reaction.mdx
@@ -47,10 +47,10 @@ dispose();
 
 ## reaction
 
-### `ReactionDisposer reaction<T>(T Function(Reaction) predicate, void Function(T) effect)`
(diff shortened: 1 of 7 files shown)


pavanpodila:
> Sidenote: suggest agreeing on a standard in the API docs. I noticed there are cases where functions are referred to as `<function>` and other places use `<function>()`. I would suggest the former and it's also the approach used in the Flutter docs

Agree with the consistency of style. We can do it the way Flutter docs do it. Sounds like a new PR :-)

PR #399, opened 2020-01-16
Fix @observable generic types when using import aliases

### LINKED ISSUE #367: observable list isn't a valid override of
After update the mobx version I'm having issue with all of my Stores that have a observable list. Lets say that I have the following code:

```dart
class EventsStore = _EventsStore with _$EventsStore;

abstract class _EventsStore with Store {

  final EventsService eventsService;

  _EventsStore(this.eventsService) : assert(eventsService != null);

//...

  @observable
  List<Event> _events = <Event>[];

//...
}
```

I'm receiving a error saying:

`error: '_$EventsStore._events' ('dynamic Function()') isn't a valid override of '_EventsStore._events' ('List<Event> Function()'). `

If I remve the `_events` property and run `flutter run build_runner` again the error disappear and it goes back to normal

shyndman:
There were a number of bugs with the previous implementation of the `LibraryScopedNameFinder`. This resolves them, as well as ensures that a single code path is followed whether or not the analyzed source code contains named imports, reducing the potential for future bugs.

The following bugs have been corrected when using named imports:
* Missing type arguments on classes
* Missing type arguments on function typedefs
* Missing prefixes from imported typedefs
* Missing prefixes from implicit type argument bounds

Fixes #367

### DIFF
--- a/mobx_codegen/lib/src/type_names.dart
+++ b/mobx_codegen/lib/src/type_names.dart
@@ -1,7 +1,5 @@
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/type.dart';
-// ignore: implementation_imports
-import 'package:analyzer/src/dart/resolver/scope.dart';
 import 'package:mobx_codegen/src/template/comma_list.dart';
 
 /// Determines the names of types within the context of a library, determining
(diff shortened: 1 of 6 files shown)


pavanpodila:
published as `0.4.1`

PR #407, opened 2020-01-23
Added Observer Listener Widget for business logic

jacobaraujo7:
This widget is basically a reaction that allows the user to execute a SnackBar, Dialog or something else that depends on the construction in the view.

It is similar to the BlocListener used in the flutter_bloc package.

Exemple:

```dart
            ObserverListener(
                listener: (_) {
                  _scaffold.currentState.showSnackBar(SnackBar(
                    content: Text('Click ${store.value}'),
                  ));
                },
                child: RaisedButton(
                  child: Text('Call Snackbar'),
                  onPressed: () {
                    store.setValue();
                  },
                ),
```


On365:
Very welcome addition to mobx. Thanks @jacobaraujo7

savioserra:
I like this proposal. But what if the user don't want an `autorun` reaction, but a `when` reaction instead? How could we achieve this? Why don't we just pass a `reaction` argument (instead of listener callback), and let the user decide?

jacobaraujo7:
@savioserra Perhaps as a widget it has to be as simple and clean as possible. A structure like when would not be interesting because of the constant rebuild of the view.

savioserra:
Yes, The `when` use case was just an example, which I think could be useful at some point. What I'm trying to say is that, in my humble opinion, we could provide some inversion of control here. You see, in this proposal we are not just using `autorun`, but we are also using an async scheduler (by using the `delay` argument). Am I missing something? Why is it required?

Also, I'm concerned about the `__initAutorun()` being called in each `build()`. I think this should be placed inside the `didChangeDependencies` method.

jacobaraujo7:
@savioserra  he is on Build because of the Hot Reload, if he stays on didChangeDependecies he will not change.

pavanpodila:
I don't think this belongs inside MobX. I would suggest making a separate `mobx_widgets` package that can contain a suite of such utilities and widgets.

jacobaraujo7:
I will create the package.  any suggestions for the initial versioning?

pavanpodila:
> I will create the package. any suggestions for the initial versioning?

You can start with `1.0` as your first version :-) and keep it outside this repo. I suggest creating a separate repo for `flutterando` and moving it under that?

PR #414, opened 2020-01-28
Add generation of toString method when @tostring annotation is used

hawkbee1:
Be careful: pubscpec.yaml contain dependency_overrides in order to
expose @tostring annotation from new mobx version !

Remove @store annotation from mobx lib and test
Add @tostring annotation in mobx lib and test
Add @tostring annotation recognition in mobx_codegen
Add toString method generation on computed and observables when
@tostring is used
Add test on toString method generation

Feature request: mobx_codegen: allow generating toString method for
Store class mobxjs#174
Previous pull request closed by error:
mobxjs#391

Sorry for keeping all my local history details :(


pavanpodila:
I am thinking if we even need an annotation for this. Can't `toString()` just be there by default. There is always an option to override in the outer public class. Curious if there are any issues in that approach.

@shyndman, @rrousselGit ?

rrousselGit:
I wouldn't like this to be the default, because, in the end, the goal should be to use [debugFillProperties](https://api.flutter.dev/flutter/foundation/Diagnosticable/debugFillProperties.html) instead

Using this instead of a simple `toString` will give a much better visualisation of the content inside devtools.

Here's an example:

Screenshot 2020-01-22 at 03 56 06

rrousselGit:
The main issue is that `debugFillProperties` and a classical `toString` are somehow incompatible since the former overrides the prototype of `toString` to add some optional parameters.

mobx_codegen/lib/src/template/store.dart
```diff
@@ -56,6 +60,30 @@ abstract class StoreTemplate {
 
   $actions''';
 
+    if (generateToString) {
+      for(var i = 0; i < observables.templates.length; i++){
```
pavanpodila: Why not use a map instead of an imperative loop ?

```dart
observables.templates.map ((current){ /*... */ })
```

mobx_codegen/lib/src/template/store.dart
```diff
@@ -56,6 +60,30 @@ abstract class StoreTemplate {
...
+        final ObservableTemplate current =  observables.templates[i];
+        toStringList.add('${current.name}: \${${current.name}.toString()}');
+      }
+
+      for(var i = 0; i < computeds.templates.length; i++){
```
pavanpodila: `.map` here too

mobx_codegen/lib/src/template/store.dart
```diff
@@ -56,6 +60,30 @@ abstract class StoreTemplate {
...
+        final ComputedTemplate current =  computeds.templates[i];
+        toStringList.add('${current.name}: \${${current.name}.toString()}');
+      }
+
+      toStringMethod = '''
+  @override
+  String toString() {
+    final toStringList = [${toStringList.sublist(1).fold("'${toStringList[0]}'", (curr, next) => "$curr, '$next'")}];
+    final string = toStringList.reduce((curr, next) => '\$curr, \$next');
```
pavanpodila: This could be done with a `toStringList.join(', ')` instead

shyndman:
> I am thinking if we even need an annotation for this. Can't `toString()` just be there by default. There is always an option to override in the outer public class. Curious if there are any issues in that approach.
> 
> @shyndman, @rrousselGit ?

For me, it just feels annoying. It means I have to do more work than I want to, because I want my own toStrings.

Because it isn't as simple as overriding in the public class. I'd have to copy all my non-factory constructors over to the public class as well, which is a requirement as soon as the public class has a body (as far as I can tell). The public/private/mixin split is already verbose enough, now I'm doubling constructors. argh!

But here's what wouldn't annoy me — if I already have my own toString defined in the private class, do nothing.

pavanpodila:
@shyndman, I agree on less work and would want it by default. If we take that route, we should have a `@skipToString` annotation. @hawkbee1, can we make that change ?

shyndman:
And longer-term, I'm with @rrousselGit. I'd love to see `debugFillProperties` support. This issue needs upvoting: https://github.com/flutter/flutter/issues/48511

> @shyndman, I agree on less work and would want it by default. If we take that route, we should have a `@skipToString` annotation. @hawkbee1, can we make that change ?

Before you do that, what do you think of this suggestion @pavanpodila?

> But here's what wouldn't annoy me — if I already have my own toString defined in the private class, do nothing.

pavanpodila:
> But here's what wouldn't annoy me — if I already have my own toString defined in the private class, do nothing.

Ya, I agree. We should do that

mobx_codegen/lib/src/template/store.dart
hawkbee1: done

mobx_codegen/lib/src/template/store.dart
hawkbee1: done

mobx_codegen/lib/src/template/store.dart
hawkbee1: done

hawkbee1:
> @shyndman, I agree on less work and would want it by default. If we take that route, we should have a `@skipToString` annotation. @hawkbee1, can we make that change ?
> 
> BTW, I am not married to that annotation name, we should pick something that is intuitive. The other idea I had was to go with something verbose like:
> 
> `@storeConfig(toString: false)`. This makes it more future-proof for further additions later

Both possibilities are easy implement.

hawkbee1:
> And longer-term, I'm with @rrousselGit. I'd love to see `debugFillProperties` support. This issue needs upvoting: [flutter/flutter#48511](https://github.com/flutter/flutter/issues/48511)

I now have some doubt or misunderstanding. debugFillProperties looks specific to StefulWidget so I don't see the point of removing it from flutter.
Shouldn't we create a specific field annotation ?

I don't understand when toString method implementation works fine using DiagnosticsProperty directly.

rrousselGit:
> I now have some doubts or misunderstanding. debugFillProperties looks specific to StatefulWidget, so I don't see the point of removing it from flutter.

It's not specific to StatefulWidget.
Flutter has tons of classes implementing Diagnosticable

It's basically the contract used by the devtools to convert _any_ class into something an expandable graph of properties

shyndman:
> > @shyndman, I agree on less work and would want it by default. If we take that route, we should have a `@skipToString` annotation. @hawkbee1, can we make that change ?
> > BTW, I am not married to that annotation name, we should pick something that is intuitive. The other idea I had was to go with something verbose like:
> > `@storeConfig(toString: false)`. This makes it more future-proof for further additions later
> 
> Both possibilities are easy implement.

Then can we have both? It'd be nice not to blast people's code with toStrings on the next upgrade.

rrousselGit:
What about being able to configure the default inside the `build.yaml`?

rrousselGit:
Here's an example on a generator of mine: https://github.com/rrousselgit/functional_widget#customize-the-output

mobx/lib/mobx.dart
```diff
@@ -30,7 +30,7 @@ library mobx;
 
 export 'package:mobx/src/api/action.dart';
 export 'package:mobx/src/api/annotations.dart'
-    show action, computed, observable, store;
+    show action, computed, observable, toString;
```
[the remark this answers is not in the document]
hawkbee1: I don't see how I can create a storeConfig(toString: false) constant.
I just used @MakeStoreConfig(makeToString: true) with default to false in order to prevent bad surprises for users.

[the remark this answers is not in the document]
hawkbee1: toString is an override so I named the parameter makeToString.

mobx/lib/mobx.dart
```diff
@@ -30,7 +30,7 @@ library mobx;
 
 export 'package:mobx/src/api/action.dart';
 export 'package:mobx/src/api/annotations.dart'
-    show action, computed, observable, toString;
+    show action, computed, observable, MakeStoreConfig;
```
pavanpodila: The annotation should be called `storeConfig` and exported as such. `MakeStoreConfig` is only the type of the annotation

mobx/lib/src/api/annotations.dart
```diff
@@ -1,14 +1,11 @@
...
-class MakeToString {
-//  unnamed constructor to allow parameters in Store annotation and prevent useless code refactoring for mobx users
-  const MakeToString._();
+class MakeStoreConfig {
+  const MakeStoreConfig({this.makeToString = false});
+  final bool makeToString;
 }
 
-/// Declares the annotation identifier to generate toString() on observables and computeds.
-const MakeToString toString = MakeToString._();
```
pavanpodila: Need to create an instance for `storeConfig`:

```dart
const MakeStoreConfig storeConfig = MakeStoreConfig._();
```

mobx/lib/src/api/annotations.dart
```diff
@@ -1,14 +1,11 @@
 /// Internal class only used for code-generation with `mobx_codegen`.
 ///
 /// During code-generation, this type is detected to identify Store class with generated toString() method
-class MakeToString {
-//  unnamed constructor to allow parameters in Store annotation and prevent useless code refactoring for mobx users
-  const MakeToString._();
+class MakeStoreConfig {
+  const MakeStoreConfig({this.makeToString = false});
```
pavanpodila: field should be called `toString` instead of `makeToString`

mobx_codegen/test/mobx_codegen_test.dart
```diff
@@ -34,8 +34,8 @@ void main() {
       await compareFiles('./data/valid_input.dart', './data/valid_output.dart');
     });
 
-    test('generates for a class mixing Store with annotation @MakeToString', () async {
-      await compareFiles('./data/valid_input_annotation_make_to_string.dart', './data/valid_output_annotation_make_to_string.dart');
+    test('generates for a class mixing Store with annotation @StoreConfig(toString: true)', () async {
```
pavanpodila: names should be consistent everywhere, esp. for `storeConfig`

mobx/lib/src/api/annotations.dart
hawkbee1: > toString is an override so I named the parameter makeToString.
When using toString everything is falling apart.

mobx/lib/src/api/annotations.dart
hawkbee1: Of course but I don't see how we add the parameter part like this. Would you be ok with:
```dart
class StoreConfig extends MakeStoreConfig {
  const StoreConfig({this.makeToString = false});
  const StoreConfig._({this.makeToString = false});
  @override
  final bool makeToString;
}
```

mobx/lib/mobx.dart
hawkbee1: You mean StoreConfig to deal with parameter parts ?

mobx_codegen/test/mobx_codegen_test.dart
hawkbee1: oups

mobx_codegen/test/mobx_codegen_test.dart
```diff
@@ -34,6 +34,10 @@ void main() {
       await compareFiles('./data/valid_input.dart', './data/valid_output.dart');
     });
 
+    test('generates for a class mixing Store with annotation @StoreConfig(toString: true)', () async {
```
pavanpodila: StoreConfig is camel case: `storeConfig`. Its the annotation that end users will be using

mobx/lib/src/api/annotations.dart
pavanpodila: how about a more explicit `hasToString`? It's a boolean field and `hasXXX`, `isXXX` is standard practice.

mobx/lib/mobx.dart
pavanpodila: > You mean StoreConfig to deal with parameter parts ?

Yes, let's call it `StoreConfig`

hawkbee1:
I was about to do the pull request but, before, I'd like to tackle what maybe a misunderstanding about the annotation the end user will use.

I see 2 ways of doing things ... and we can do both but I'm afraid of confusion for end user.

1/ @storeConfig

> Need to create an instance for `storeConfig`:
> 
> ```dart
> const MakeStoreConfig storeConfig = MakeStoreConfig._();
> ```

2/ @StoreConfig(makeToString: true)
I insist on the S of StoreConfig because I see the code like this:  

> ```dart
> class StoreConfig extends MakeStoreConfig {
>   const StoreConfig({this.makeToString = false});
>   const StoreConfig._({this.makeToString = false});
>   @override
>   final bool makeToString;
> }
> ```

I'm setting makeToString default to false for compatibility purpose.

hawkbee1:
Ok we published at same time. 
I'm preparing the pull request.

pavanpodila:
Let's keep the main annotation class as `StoreConfig`, `S` capital. It will have a const constructor with `hasToString` as the field. Let's keep it `true`, since the popular vote is to have it by default

hawkbee1:
I suppose I add @StoreConfig(hasToString: false) to all existing mobx_codegen test input files ?

pavanpodila:
> I suppose I add @StoreConfig(hasToString: false) to all existing mobx_codegen test input files ?

Yes for existing files.

pavanpodila:
Also need version bump in pubspec.yaml and info in Changelog for this new feature :-). Give credit to yourself in Changelog!

hawkbee1:
I wasn't able to write "thanks to" :-)

Important reminder: pubscpec.yaml contain dependency_overrides for testing purpose !

mobx/lib/src/api/annotations.dart
```diff
@@ -1,12 +1,19 @@
...
+/// During code-generation, this type is detected to identify Store class with generated toString() method
+class MakeStoreConfig {
+  const MakeStoreConfig({this.hasToString = true});
+  const MakeStoreConfig._({this.hasToString = true});
+  final bool hasToString;
 }
 
-/// Declares the annotation identifier for a Store class.
-const MakeStore store = MakeStore._();
+class StoreConfig extends MakeStoreConfig {
```
pavanpodila: Why do we need a subclass? We can just rename `MakeStoreConfig` -> `StoreConfig`?
```suggestion
class StoreConfig {
```

mobx/lib/src/api/annotations.dart
```diff
@@ -1,12 +1,19 @@
...
 ///
-/// During code-generation, this type is detected to identify a MobX Store class
-class MakeStore {
-  const MakeStore._();
+/// During code-generation, this type is detected to identify Store class with generated toString() method
+class MakeStoreConfig {
+  const MakeStoreConfig({this.hasToString = true});
+  const MakeStoreConfig._({this.hasToString = true});
+  final bool hasToString;
 }
```
pavanpodila: Don't need this

mobx_codegen/pubspec.yaml
```diff
@@ -16,10 +16,15 @@ dependencies:
   path: ^1.6.2
   source_gen: ^0.9.4
 
+dependency_overrides:
+  mobx:
+    path: ../mobx
+
```
pavanpodila: We should remove this

mobx/lib/src/api/annotations.dart
pavanpodila:
```suggestion
class StoreConfig {
```

mobx/lib/src/api/annotations.dart
```diff
@@ -1,12 +1,19 @@
...
 
-/// Declares the annotation identifier for a Store class.
-const MakeStore store = MakeStore._();
+class StoreConfig extends MakeStoreConfig {
+  const StoreConfig({this.hasToString = true});
+  const StoreConfig._({this.hasToString = true});
```
pavanpodila:
```suggestion

```

mobx/lib/src/api/annotations.dart
```diff
@@ -1,12 +1,19 @@
...
+  @override
```
pavanpodila:
```suggestion

```

mobx/lib/src/api/annotations.dart
```diff
@@ -1,12 +1,19 @@
...
+  // ignore: overridden_fields
```
pavanpodila:
```suggestion

```

mobx_codegen/lib/src/store_class_visitor.dart
```diff
@@ -4,7 +4,7 @@ import 'package:build/build.dart';
 import 'package:mobx/mobx.dart';
 // ignore: implementation_imports
 import 'package:mobx/src/api/annotations.dart'
-    show ComputedMethod, MakeAction, MakeObservable;
+    show ComputedMethod, MakeAction, MakeObservable, MakeStoreConfig;
```
pavanpodila:
```suggestion
    show ComputedMethod, MakeAction, MakeObservable, StoreConfig;
```

mobx_codegen/pubspec.yaml
hawkbee1: yes, this one was just to let the tests run correctly during dev

pavanpodila:
> I wasn't able to write "thanks to" :-)

You can do it 👍

mobx_codegen/pubspec.yaml
pavanpodila: > yes, this one was just to let the tests run correctly during dev

Agree. We can also publish to `.alpha` versions for testing purposes

mobx/lib/mobx.dart
```diff
@@ -29,8 +29,8 @@
 library mobx;
 
 export 'package:mobx/src/api/action.dart';
-export 'package:mobx/src/api/annotations.dart'
-    show action, computed, observable, store;
+export 'package:mobx/src/api/annotations.dart';
+//    show action, computed, observable, StoreConfig;
```
pavanpodila: We need to uncomment this line. The reason we have this is not to expose `Make**` classes, hence the "show statement"

hawkbee1: Sorry, I forgot this one.

mobx_codegen/lib/src/template/store.dart
```diff
@@ -56,6 +60,25 @@ abstract class StoreTemplate {
 
   $actions''';
 
+    if (generateToString) {
+      toStringList
+        ..addAll(observables.templates.map(
+            (current) => '${current.name}: \${${current.name}.toString()}'))
+        ..addAll(computeds.templates.map(
+            (current) => '${current.name}: \${${current.name}.toString()}'));
```
pavanpodila: We should filter to only include fields that are **public**.

mobx_codegen/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.0.2
+
+- Added automatic generation of toString method with @StoreConfig annotation
```
pavanpodila: Enclosing in backticks for code formatting.

```suggestion
- Added automatic generation of `toString` method with `@StoreConfig` annotation 
```

mobx_codegen/lib/src/store_class_visitor.dart
```diff
@@ -189,10 +190,24 @@ class StoreClassVisitor extends SimpleElementVisitor {
...
+const _toStringAnnotationChecker = TypeChecker.fromRuntime(StoreConfig);
 
 bool isMixinStoreClass(ClassElement classElement) =>
     classElement.mixins.any(_storeMixinChecker.isExactlyType);
 
+// Checks if the class as a toString annotation
+bool isToStringAnnotatedStoreClass(ClassElement classElement) =>
+    _toStringAnnotationChecker.hasAnnotationOfExact(classElement);
+
+bool isGenerateToStringTrue(ClassElement classElement) {
```
pavanpodila: Let's call this `hasGeneratedToString`

mobx_codegen/lib/src/template/store.dart
hawkbee1: you mean removing all fields with name starting with '_'  ?

pavanpodila: Yes, but the check should be on the element, not the field name. I think you can check the access level of the element (`isPublic`??) and then filter.

hawkbee1: At this point it's too late to use isPublic.
I can check isPublic when we add elements to observables list and computed list. At this point I can either mark the element as public in its template or add it to a specific toString list inside the StoreTemplate.

pavanpodila: no worries, take your time. Yes, we should test before adding the fields to the list. We can check for `isPublic` before adding. The PR is shaping up well. Great job so far! We are close to the finish line

hawkbee1:
I chose to ad isPrivate in observable and computed template. 
I'm not certain creating a private computed getter make sense but.. better safe than sorry.

mobx_codegen/lib/src/template/store.dart
```diff
@@ -61,10 +61,16 @@ abstract class StoreTemplate {
   $actions''';
 
     if (generateToString) {
+      final publicObservablesList = observables.templates
+      ..removeWhere((element) => element.isPrivate);
+
+      final publiccomputedsList = computeds.templates
```
pavanpodila: camel casing is off. Should be `publicComputedsList`, 'C' capitalized

pavanpodila:
Also there are some analyzer errors, as can be seen in the failed build steps

mobx_codegen/pubspec.yaml
hawkbee1: How to proceed with publishing .alpha version ?

hawkbee1:
The fail in mobx_codgen is due to the annotation in mobx. I have the same if I remove my dependency override on local.
 I will check later for the fail in flutter_mobx.

mobx_codegen/lib/src/template/store.dart
hawkbee1: :-(

pavanpodila:
> I chose to ad isPrivate in observable and computed template.
> I'm not certain creating a private computed getter make sense but.. better safe than sorry.

You did the right thing. Both can be private :-)

pavanpodila:
> The fail in mobx_codgen is due to the annotation in mobx. I have the same if I remove my dependency override on local.
> I will check later for the fail in flutter_mobx.

Some of the errors seem to be pure analyzer errors

hawkbee1:
The error in flutter_mobx is pure analyzer error ?

We have to wait for extraction of diagnostics before working on debugFillProperties. I see @rrousselGit  is working on the subject.

pavanpodila:
I think we need to split this into 2 PRs:

1. Publish StoreConfig annotation from mobx
2. Consume in mobx_codegen for toString

We already most of 2., just need to make a separate PR for 1.

hawkbee1:
Tests on the separate PR are looking better :dancer:

mobx_codegen/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: mobx_codegen
 description: Code generator for MobX that adds support for annotating your code with @observable, @computed, @action and also creating Store classes.
-version: 1.0.1
+version: 1.0.2
```
pavanpodila: We are almost done here. Just need to add the entry in the changelog for `mobx_codegen`

hawkbee1: I added myself in the contributors list :-)

pavanpodila:
Thanks @hawkbee1 for your time and patience in getting this useful feature merged 🎉

Issue #416, opened 2020-01-30
ObservableMap doesn't set null values


Vardiak:
If we add `null` as a new value in an ObservableMap, the map isn't updated.

```
final map = ObservableMap();
map['test'] = null;
print(map);
```

Here, we can see that the map is still `{}`.
It works if we update the value after being set.

```
final map = ObservableMap();
map['test'] = 'value';
map['test'] = null;
print(map);
```

I found where this might come from, I will submit a pull request later today.

pavanpodila:
The code indicates that you are using a plain map. Hope you are using an instance of the `ObservableMap`?

Vardiak:
Sorry I messed up my example, yes it works when using plain map, but not when using an ObservableMap.

pavanpodila:
Fixed and merged

PR #417, opened 2020-01-30
Fix ObservableMap not adding null values

Vardiak:
Adressing #416

### DIFF
--- a/mobx/lib/src/api/observable_collections/observable_map.dart
+++ b/mobx/lib/src/api/observable_collections/observable_map.dart
@@ -86,7 +86,7 @@ class ObservableMap<K, V>
         }
       }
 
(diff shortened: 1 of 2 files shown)


mobx/lib/src/api/observable_collections/observable_map.dart
```diff
@@ -86,7 +86,7 @@ class ObservableMap<K, V>
         }
       }
 
-      if (value != oldValue) {
+      if (!_map.containsKey(key) || value != oldValue) {
```
pavanpodila: Can you add a quick test as well for this code path?

pavanpodila:
Thanks! Its now published in mobx 1.0.1

PR #421, opened 2020-02-05
Corrected ObservableMap documentation

geisterfurz007:
The documentation incorrectly stated that this class is the observable counterpart to the standard dart List interface while it is the counterpart to the Map interface.


pavanpodila:
Thanks! Added you as a contributor :-)

PR #435, opened 2020-02-13
docs on linking stores and best practices when working with mobx

pavanpodila:


irvine5k:
I don't know if you are done with Linking Stores section but linking Stores using reactions like BLoC recommendations is a good option?

IMHO it can mess our UI code and when you're testing this link you'll need to make a test widget instead of unit test on Stores.

pavanpodila:
Do you have a link to that example?

rullyalves:
This image shows a communication between a login store and an authentication store, being made based on the user's user interface, which would be a bad practice? in my view, communicating Stores directly (calling a store within another can generate an open coupling).

image

irvine5k:
> Do you have a link to that example?

image

Something like this, If I want to test it, i'll need to make a test widget. This example is so simple, but it could envolve a lot of logic.

What's is your opinion about this approach?

I don't know if you saw my edit on the first comment, are you going to write about navigation/alerts/snackbars based on reactions?

rullyalves:
@irvine5k, @pavanpodila 
It makes me think that maybe the interaction in Stores should be done on a facade, in order to decouple that from the widgets

image

pavanpodila:
> I don't know if you saw my edit on the first comment, are you going to write about navigation/alerts/snackbars based on reactions?

All UI related behaviors should be handled inside the StatefulWidget. These don't really belong to the Store, so its best to setup these reactions inside `initState` and then dispose later. The reactions can monitor fields from your store, which can be set before showing the alert/navigation/snackbar. Hope that makes sense.

PR #440, opened 2020-02-23
Replace json_serializable by dart_json_mapper in Examples module

k-paxian:


deadsoul44:
I love dart_json_mapper but it would be good to have at least one example with json_serializable to remind developers that they have another option also.

k-paxian:
@deadsoul44 

image

Just followed suggestion from @pavanpodila

deadsoul44:
Totally acceptable. Great work. 
I hope some people will add examples with json_serializable and built_value maybe also freezed in future.
Thank you.

mobx_examples/lib/connectivity/connectivity_store.dart
```diff
@@ -1,12 +1,8 @@
-import 'package:mobx/mobx.dart';
 import 'package:connectivity/connectivity.dart';
+import 'package:mobx/mobx.dart';
 
-part 'connectivity_store.g.dart';
```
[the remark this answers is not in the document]
k-paxian: Reverted g.part code.
The funny thing is that w/o that code examples are working the same way as with it.
No fallen tests, no red errors in console, that makes me thinking it's not a vital files 😄

pavanpodila:
@all-contributors add @k-paxian for example

mobx_examples/.gitignore
```diff
@@ -8,6 +8,7 @@
 .buildlog/
 .history
 .svn/
+*.reflectable.dart
```
MaikuB: @pavanpodila are you ok if we revert this. Was taking a look at the examples again and I couldn't run it after the packages were downloaded as this rule meant code generation needs to be triggered. Off memory, this is something we had tried to avoid and we've been committing the generated code for the stores so that users could get the examples up and running sooner

pavanpodila: Yes, let's change it! PR ?

MaikuB: Yep I can do that within the next couple of days. Was looking to bump the provider dependency, which is why I was looking at the examples :)

PR #451, opened 2020-03-12
[feat] spying

pavanpodila:
Adds spy support to MobX. This allows:

- Setting up spies to listen to activity happening inside MobX, such as observables getting updates, computed getting reevaluated, reactions firing, actions firing, etc.
- All activities are now tracked
- In some cases proper names were not propagated in classes like `ObservableFuture`, `ObservableStream`, `ActionController`, which has been fixed
- `mobx_codegen` related changes that brings the spying support to generated getters/setters/actions


rfc/spying-and-tracing.md
```diff
@@ -0,0 +1,24 @@
+# Spying and Tracing
+
+This feature adds the ability to expose the inner workings of MobX. It will report over console logs or DDT about the various activities happening inside MobX. This includes:
+
+- When observables are read and written
+- When actions are invoked
+- When reactions are triggered
+
+## Benefits
```
shyndman: I think this section could use more flavor. Maybe problems that we'd like the feature to solve for people? Use cases? More detail can help shape the feature and ensure that we're solving real user needs.

pavanpodila: Agree. Will add more details here and how this feature helps

shyndman: I'll help. I have a few things in mind, but they're still brewing. :)

pavanpodila: Pushed a few changes. Let me know if this is along those lines :-)

pavanpodila:
@shyndman can you review? I know its a lot of files but many of them have  few lines of change, mostly to fire the spyEvent and propagate names to internal structures.

I am a bit stumped with the build failures. All tests pass locally and coverage is up to **98.8%**, but not sure why they fail on CI. Any ideas?

PR #462, opened 2020-04-04
bump dependencies used by mobx_examples, fix settings example and add reflectable code

MaikuB:
note that I saw a note on the reflectable package on not committing the generated code. I could revert that as it seems like it's still a subjective matter and the `android` and `ios` folders aren't in source control (although `web` is) so looks like we need users to run commands to get the examples up and running anyway


MaikuB:
@pavanpodila i'm not sure what's going on with the build failures but one thing I did notice is the widget test fails when run locally. Not sure if it's part of CI right now but may want to delete it as it's currently not relevant in that it passes the app to do a widget test but the tests seem to be the one that comes out of the box for the counter app

pavanpodila:
Ya not sure why the mobx tests are failing. They do pass locally. My other PR(#451) is also having the same issue where the `test_coverage` package fails internally.

MaikuB:
Sorry if I wasn't clear. There's actually a widget test that is part of `mobx_examples` that is failing. I don't think it's meant to be there and suggest deleting it. Thoughts on that?

pavanpodila:
> Sorry if I wasn't clear. There's actually a widget test that is part of `mobx_examples` that is failing (i.e. it's a separate thing to CI failing). I don't think it's meant to be there and suggest deleting it. Thoughts on that?

ya let's delete that. Also mobx_examples are not under CI

MaikuB:
Yep i figured as much as it was already failing before I made changes. I've pushed some changes. if you could, it'd be great if you can try running the Github example as it keeps returning a resource not found error when I trigger a search. Was working for me before so not sure what's going on

MaikuB:
Double checked and have the same error with the GitHub example when i switch to master branch. Perhaps an auth related issue although the error doesn't indicate it as such

PR #463, opened 2020-04-04
Fix a typo in Computed Observables section

polilluminato:
I think that in the example of Computed Observables the class Contact as part with name "contact.g.dart" and not "counter.g.dart".
Maybe it's a typo from the previuos example.


pavanpodila:
@all-contributors add @polilluminato  for docs

PR #469, opened 2020-04-07
Document translation of zh-CN.

bsdfzzzy:
Hi all,
Thanks for the wonderful tool. It helps me a lot in flutter. I noticed that there is no Chinese doc. So I spend some time to do that. Hopes can help more people interested in mobx.dart.


pavanpodila:
@all-contributors add @bsdfzzzy for doc

PR #470, opened 2020-04-07
Add map function on Future extension

LucazzP:
Transform the future value into another type after complete future. Useful to transform List to ObservableList in requests

### DIFF
--- a/mobx/lib/src/api/extensions/observable_future_extension.dart
+++ b/mobx/lib/src/api/extensions/observable_future_extension.dart
@@ -4,4 +4,7 @@ part of '../extensions.dart';
 extension ObservableFutureExtension<T> on Future<T> {
   ObservableFuture<T> asObservable({ReactiveContext context, String name}) =>
       ObservableFuture<T>(this, context: context, name: name);
+  /// Transform the future value into other type after complete future
+  /// Useful to transform list to ObservableList in requests
+  Future<E> map<E>(Future<E> Function(Future<T>) transform) => transform(this);
 }


mobx/lib/src/api/extensions/observable_future_extension.dart
```diff
@@ -4,4 +4,7 @@ part of '../extensions.dart';
 extension ObservableFutureExtension<T> on Future<T> {
   ObservableFuture<T> asObservable({ReactiveContext context, String name}) =>
       ObservableFuture<T>(this, context: context, name: name);
+  /// Transform the future value into other type after complete future
```
pavanpodila, line 7: I feel this is more of an application level feature than a core-need at the MobX layer. We had this discussion on Discord where such extensions should be in a separate `mobx-utils` or some other community driven package. Once we hone in some core set of patterns and APIs, we could then pull them into mobx-core. For now, I suggest we start with `mobx-community` or `mobx-utils` and put this over there.


What do you think?

LucazzP: I think is a good idea to separate those things, but the extensions have a "problem" that doesn't work with auto import. If I use the package how is now, I don't need to import the extensions package, such as the `mobx-utils` how you proposed,  because the mobx package already was imported on the controller class. So, if we transfer those things that aren't the core of mobx, we will need to manually write the import on the controller to use the extensions. 
But of course, my feature is an util feature and the best practice is separate this in a utils package, but we need to think more to not difficult more the life of the developer.
So, an opinion is not create another package, but use the `flutter_mobx` package, because have this same propose that you want, an util package.

pavanpodila: `flutter_mobx` is serving a single need of providing the glue to connect a mobx Store to Flutter. It is quite limited in that sense and should not be doing anything more.

pavanpodila: The reason I want to keep utilities separate is to keep the core lean, else it is easy to keep piling more things as being "essential", when in fact they would be very use case driven.

LucazzP: Ok, it makes sense...

pavanpodila:
Closing this for now as this extension is at the application level

PR #479, opened 2020-04-23
Add a throwReactionErrors setup task.

hmayer00:
This will allow expectations in observers to function.

Currently, test expectations written inside of observation handlers don't function. For example, try changing the integer used in the expectation around line 90 in computed_test.dart. Since expectation failures are implemented as a thrown exception, and reaction errors are swallowed unless there is an `onReactionError` handler, tests appear to pass even if the expectations inside of observers fail.

Probably this method call should become standard boilerplate, but I haven't added it to any files other than the one I was looking at while doing a separate PR (coming very shortly). I'm happy to discuss that further.

### DIFF
--- a/mobx/test/action_controller_test.dart
+++ b/mobx/test/action_controller_test.dart
@@ -9,6 +9,7 @@ class MockDerivation extends Mock implements Derivation {}
 
 void main() {
   turnOffWritePolicy();
+  testSetup();
 
   group('ActionController', () {
     test('can be created with both null context and name', () {
--- a/mobx/test/action_controller_test.dart
+++ b/mobx/test/action_controller_test.dart
@@ -8,7 +8,6 @@ import 'util.dart';
 class MockDerivation extends Mock implements Derivation {}
 
 void main() {
-  turnOffWritePolicy();
   testSetup();
 
   group('ActionController', () {
--- a/mobx/test/action_test.dart
+++ b/mobx/test/action_test.dart
@@ -7,6 +7,7 @@ import 'util.dart';
 
 void main() {
   turnOffWritePolicy();
+  testSetup();
 
   group('Action', () {
     test('basics work', () {
--- a/mobx/test/action_test.dart
+++ b/mobx/test/action_test.dart
@@ -6,7 +6,6 @@ import 'shared_mocks.dart';
 import 'util.dart';
 
 void main() {
-  turnOffWritePolicy();
   testSetup();
 
   group('Action', () {
--- a/mobx/test/annotations_test.dart
+++ b/mobx/test/annotations_test.dart
@@ -1,7 +1,11 @@
 import 'package:mobx/mobx.dart';
 import 'package:test/test.dart';
 
+import 'util.dart';
+
 void main() {
+  testSetup();
+
   test('annotations are defined', () {
     expect(observable, isNotNull);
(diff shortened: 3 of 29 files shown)


pavanpodila:
How about just using the `onError` parameter for reactions? I understand this is to catch errors everywhere and not have to setup for each Reaction or ones which are deeper like the `reaction` inside an `Observer`?

hmayer00:
Well, really it's just so that expectations inside of reactions work. I was working on the custom `equals` method for Computed, and being test-driven about it. So I wrote the test first, expecting it to fail, and it passed. Took me a while to realize that it's because any expectation written inside an observation will look like it's succeeding. So some of the expectations that are already in `computed_test.dart` are worthless at the moment. This seemed like the most straightforward way to fix that, rather than leaving that around as a timebomb for future tests.

hmayer00:
So this block around line 90 in `computed_test.dart`, for example:
```dart
final dispose1 = total.observe((change) {
  expect(change.newValue, equals(30));
  expect(executionCount, equals(1));
});
```

If you were to change the values in the `equals` calls, the tests would all still pass, using current code.

pavanpodila:
> So this block around line 90 in `computed_test.dart`, for example:
> 
> ```dart
> final dispose1 = total.observe((change) {
>   expect(change.newValue, equals(30));
>   expect(executionCount, equals(1));
> });
> ```
> 
> If you were to change the values in the `equals` calls, the tests would all still pass, using current code.

This should definitely be fixed. Can you include this in the PR as well ?

hmayer00:
> This should definitely be fixed. Can you include this in the PR as well ?

Oh, sorry, I am definitely not doing a good job explaining myself. That is the PR! Calling throwReactionErrors as a setup task will mean that all those expectations start working. 

There are probably other ways to go about it if you're not a fan of the execution. For example we could check to see if the exception is a `TestFailure`, but in my limited testing that also got wrapped in a `MobXCaughtException` and then I had to do something like `rxn.errorValue.exception is TestFailure || rxn.errorValue.exception.message.exception is TestFailure` and the whole thing seemed like a fragile mess. So this seemed like the simpler solution.

Also my hunch is something like this ought to be included in other (perhaps all?) test files. But I dunno if it would cause other issues.

pavanpodila:
Got it! Why don't we try including this in few other files and see if something fails? I think these are global guard rails, which we ought to have for reliable testing. Thanks for catching this!

hmayer00:
Yeah looks like that should work for the most part, though there are some files where throwing reaction errors breaks things. I have a good-enough solution for now I think. Gotta run off for a bit but should be able to push it in a few hours.

hmayer00:
Alright, in one commit I added a `testSetup` function that sets up `throwReactionErrors` as an opt-out, and called that function in every test file (opting out in 2 files). In another commit I included `turnOffWritePolicy` in that same `testSetup` function, also as an opt-out, though as far as I can tell it did no harm in the files that didn't already include it, so I didn't opt out anywhere.

It's definitely imperfect, since the files that opt out of throwReactionErrors are vulnerable to the original problem. But I don't see a better solution right now, and I don't really have the bandwidth to dig in further. Certainly the changes in this PR would leave the general state of things better off, though there's still a potential problem lurking.

pavanpodila:
@allContributors add @hmayer00 for code, test

PR #480, opened 2020-04-23
Fire computed observers when the computed changes.

hmayer00:
Make `Computed.observe` actually call the handler when the computed changes.

I could definitely use some help on this one.

From my attempts to understand what's happening here, it looks like `observe` on `Computed` is basically nonfunctional. The handler is called when the Computed is initialized, because it is the first time. And then after that it will never be called again, unless it was initialized with `fireImmediately: true`. `fireImmediately` has no impact on whether it fires immediately, since it being the first time is enough to accomplish that. But it does determine whether the handler is ever called again. This seems clear from the code, and both automated testing and manual testing agree.

On the other hand it's hard to believe that this could be true, since one would think that would cause immediate bugs for anyone who used a Computed. I'm about to get pretty over my head since I don't really understand how mobx works under the hood, but I think that it doesn't come up much because the Observer flutter widget doesn't use `Computed.observe`, but is notified in other ways? And perhaps in various reactive contexts it might wind up being sort of invisible because if you initialize the Computed within that context, it will access the underlying observables, and then be notified anyway, even if the observe call didn't work? Or perhaps observe is just rarely used. I dunno, I would have expected this to be a problem but since I don't see any issues about it it can't be doing that much harm. But I can definitely reproduce it in my own application (though it didn't come up naturally).

Anyway, a couple of questions I have:
- is the if/else untracked at line 167 correct? I'll leave the more specific question inline there.
- is this a breaking change? The deprecation of `fireImmediately` shouldn't do too much harm I don't think, especially since it was misleading to begin with. But are there people whose code unintentionally relied on observers not firing? Anyway, that's over my pay grade but I wanted to call it out for consideration.

### DIFF
--- a/mobx/lib/src/core/computed.dart
+++ b/mobx/lib/src/core/computed.dart
@@ -146,21 +146,28 @@ class Computed<T> extends Atom implements Derivation, ObservableValue<T> {
 
   bool _isEqual(T x, T y) => x == y;
 
+  // [fireImmediately] never did quite what it sounded like, and the handler has always
+  // fired immediately. So it is deprecated and now non-functional.
   Function observe(void Function(ChangeNotification<T>) handler,
-      {bool fireImmediately}) {
+      {@deprecated bool fireImmediately}) {
     var firstTime = true;
     T prevValue;
 
     return autorun((_) {
       final newValue = value;
-      if (firstTime == true || fireImmediately == true) {
-        _context.untracked(() {
-          handler(ChangeNotification(
-              type: OperationType.update,
-              object: this,
-              oldValue: prevValue,
-              newValue: newValue));
-        });
+
+      void handleChange() {
+        handler(ChangeNotification(
+            type: OperationType.update,
+            object: this,
+            oldValue: prevValue,
+            newValue: newValue));
+      }
+
+      if (firstTime == true) {
+        _context.untracked(handleChange);
+      } else {
+        handleChange();
       }
 
       firstTime = false;
--- a/mobx/test/computed_test.dart
+++ b/mobx/test/computed_test.dart
@@ -78,6 +78,32 @@ void main() {
       final x = Observable(10);
       final y = Observable(20);
 
+      var observationCount = 0;
+
+      final total = Computed(() => x.value + y.value);
+
+      var expectedComputedValue = x.value + y.value;
+
+      final dispose = total.observe((change) {
+        observationCount++;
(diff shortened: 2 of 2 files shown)


mobx/lib/src/core/computed.dart
```diff
@@ -146,21 +146,28 @@ class Computed<T> extends Atom implements Derivation, ObservableValue<T> {
...
+            type: OperationType.update,
+            object: this,
+            oldValue: prevValue,
+            newValue: newValue));
+      }
+
+      if (firstTime == true) {
+        _context.untracked(handleChange);
+      } else {
+        handleChange();
```
hmayer00, line 170: So I'm not clear on what work `untracked` is doing here. For some reason it felt like having the handler be untracked in the initial setup was desired, but I'm really not sure what was supposed to happen in the alternate case. This was my best guess as to the intention, but that is based on essentially nothing. So I'm certainly interested in guidance here.

pavanpodila:
TBH, `computed.observe` is rarely used. Most often folks would end up creating a `reaction` that observes the computed instead. The `observe` method was setup purely for consistency and has not been used much in practice. However that said, its worth to keep the implementation correct, so there is definitely merit in fixing this.

mobx/lib/src/core/computed.dart
pavanpodila: Since we are using `autorun`, `untracked` ensures we are not going into a self-loop because of observing  other observables invoked as part of the `handler`, which we don't care about.

pavanpodila, line 170: Even this should be inside `untracked`

pavanpodila:
I have incorporated some of the changes from the PR and also fixed a test. We can safely close this for now. The change is part of this commit: 2eee37982d5f669cec00543d66df6136ff2cd332

mobx/lib/src/core/computed.dart
hmayer00: Thanks, that makes sense.

hmayer00:
> I have incorporated some of the changes from the PR and also fixed a test. We can safely close this for now. The change is part of this commit: [2eee379](https://github.com/mobxjs/mobx.dart/commit/2eee37982d5f669cec00543d66df6136ff2cd332)

Looks great - definitely cleaner than what I had, since I was bumbling around not understanding the purpose of the untracked part. Thanks for pulling the changes in!

Issue #450, opened 2020-03-12
Consider adding an option for an equality check to @observable


shyndman:
Hello!

At the moment, `@observable`s generate setters that always set the field value and report change, whether or not the new value differs from the field's current value.

It seems like it could be useful to add an option to `@observable` to include an equality check to minimize the reporting of change.

If provided as an option, instead of a default, the change would be non-breaking.

Maybe something like:

```dart
@observable(checkEqualityOnSet: true)
String myObservableProperty;
```

`checkEqualityOnSet` is a little wordy though. Suggestions welcome.

shyndman:
Started playing with this, and it would need to be slightly different. We'd need to expose an additional annotation — one that takes args.

pavanpodila:
We have this logic on the `Observable` class, which is purely based on references/values and not structures. I think we can reuse that part here.

Reference: https://github.com/mobxjs/mobx.dart/blob/master/mobx/lib/src/core/observable.dart#L27

shyndman:
Ah, interesting. How would you recommend reusing that piece?

pavanpodila:
We can move the comparator logic to a global function that takes in a previous and new value and does equality or structural comparisons. Something like:


```dart
bool compareEqual<T>(T newValue, T oldValue) { }

```

shyndman:
I think that we might even be able to pass a function reference to the annotation, as long as it's top-level or static. I'll poke around.

pavanpodila:
I think annotations are const only constructors but do confirm

shyndman:
They are, but I believe that top-levels and statics are const values. I'll double check.

Yeah:
https://gist.github.com/shyndman/4541ffc0e8ec013dd21a8b504d2eef05

pavanpodila:
So static-fns possible right?

shyndman:
yup. static and top-level functions. I updated the gist/dartpad.

pavanpodila:
Awesome!! Road is clear then.

pavanpodila:
BTW, I was thinking of starting a RFC type README that outlines the approach for all new features we add to mobx from here on. The README will be sent in as a draft-PR and allows commenting and discussions. It will act as a spec for the implementation with a section even on the implementation strategy. This is very similar to the way dart-lang is doing it.

Worth a shot? Can we do for this feature? I also want to do this for the spying/tracing functionality.

shyndman:
Yeah, definitely. Let's do it.

[...]

PR #484, opened 2020-04-24
StoreConfig parameter to report only changed writes for an observable.

hmayer00:
Sets up a parameter in StoreConfig to turn off reporting a write if you've set an observable value that is equal to the old value. It will default to reporting the write, since that's the current behavior.

For example:
```dart
class Foo = _Foo with _$Foo;

abstract class _Foo with Store {
  @observable
  int value;
}

var foo = Foo();
foo.value = 1;
autorun((_) => print(foo.value.toString()));
foo.value = 1; // prints '1'

// ....

class Bar = _Bar with _$Bar;

@StoreConfig(reportOnEqualSet: false)
abstract class _Bar with Store {
  @observable
  int value;
}

var bar = Bar();
bar.value = 1;
autorun((_) => print(bar.value.toString()));
bar.value = 1; // doesn't print anything, because 1 == 1.
```

A few notes:
- The vast majority of the code change is to the test outputs, which should probably be ignored unless/until you're good with the rest of it.
- This does not use any custom equals from Observable, since codegen is upstream of that and I'm not really sure how we would use it.
- This mostly covers the original purpose of #450, but it doesn't address the eventual landing point, of allowing a custom equals parameter in the field-level annotation. I'm not able to bite that off right now and this feels like an easier way to get most of the value. I don't think it makes it much harder to accomplish the field-level annotation later, but it might to some extent.
- If the new value is equal to the old, it doesn't report the write, but it also doesn't set the underlying value. Probably in most cases that won't matter (the two are `==` after all), but if it does matter it wasn't clear to me whether it's better to set the new value and not report it, or not set it at all. Either could be confusing.

I think this is a useful modification, since I'd rather not have unnecessary rebuilds when something is set to an equal value. But I won't be offended if you don't like the path this took and would rather not include it.

### DIFF
--- a/mobx/lib/src/api/annotations.dart
+++ b/mobx/lib/src/api/annotations.dart
@@ -1,9 +1,13 @@
 /// Declares configuration of a Store class.
-/// Currently the only configuration used is boolean to indicate generation of toString method (true), or not (false)
+/// Currently the only configurations used are:
+/// - boolean to indicate generation of toString method (true), or not (false)
+/// - boolean to indicate whether to report writes to observables when the value is equal.
+///   Defaults to true.
 
 class StoreConfig {
-  const StoreConfig({this.hasToString = true});
+  const StoreConfig({this.hasToString = true, this.reportOnEqualSet = true});
   final bool hasToString;
+  final bool reportOnEqualSet;
 }
 
 /// Internal class only used for code-generation with `mobx_codegen`.
--- a/mobx_codegen/lib/src/store_class_visitor.dart
+++ b/mobx_codegen/lib/src/store_class_visitor.dart
@@ -62,7 +62,9 @@ class StoreClassVisitor extends SimpleElementVisitor {
           .addIf(!element.isAbstract, element.name);
     }
     // if the class is annotated to generate toString() method we add the information to the _storeTemplate
-    _storeTemplate.generateToString = hasGeneratedToString(element);
+    _storeTemplate
+      ..generateToString = hasGeneratedToString(element)
+      ..reportOnEqualSet = shouldreportOnEqualSet(element);
   }
 
   @override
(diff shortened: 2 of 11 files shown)


pavanpodila:
Equality checks should be on by default, I think. We should have to specify an flag for that. Why not turn it on by default ?

hmayer00:
> Equality checks should be on by default, I think. We should have to specify an flag for that. Why not turn it on by default ? The Observable type does this already and only reports when the value is different. In case of code_gen, we rely on atoms and this equality check was missing in it

Oh I totally agree that having the equality check is a better default, I was just trying to be non-disruptive (I never know how library owners feel about opting in or out of changes).

To clarify, you're good with this change if I add a commit switching the default behavior? I'll need to run through the test files again to work with the new default, so I just want to make sure before I do it. Also codecov is complaining but I don't feel like this has reduced coverage. Not sure how you want me to handle that.

darkstarx:
> Also codecov is complaining (about the patch, not the project), but I don't feel like this has reduced coverage. Not sure how you want me to handle that.

Hi! You can go in the codecov report details and observe the changes led to dropping the coverage (marked as red lines).
Here you are
image

hmayer00:
> Hi! You can go in the codecov report details and observe the changes led to dropping the coverage (marked as red lines).
> Here you are
> image

Sorry for the delay, life intervened. But I don't seem to be able to see that view? And it's weird since I only touched one of the highlighted lines.

When I look in codecov, I see
image

and

image

So I must be missing something, I think.

hmayer00:
> To clarify, you're good with this change if I add a commit switching the default behavior? I'll need to run through the test files again to work with the new default, so I just want to make sure before I do it. Also codecov is complaining (about the patch, not the project), but I don't feel like this has reduced coverage. Not sure how you want me to handle that.

@pavanpodila Just wanted to check in again to make sure you're good with the change broadly, so long as I change the default. Sorry for the delayed response, had some other stuff going on.

mobx/lib/src/api/annotations.dart
```diff
@@ -1,9 +1,13 @@
 /// Declares configuration of a Store class.
-/// Currently the only configuration used is boolean to indicate generation of toString method (true), or not (false)
+/// Currently the only configurations used are:
+/// - boolean to indicate generation of toString method (true), or not (false)
+/// - boolean to indicate whether to report writes to observables when the value is equal.
+///   Defaults to true.
 
 class StoreConfig {
-  const StoreConfig({this.hasToString = true});
+  const StoreConfig({this.hasToString = true, this.reportOnEqualSet = true});
```
pavanpodila, line 8: How about the param name as `reportIfEqual` ? Think it makes more sense ?

pavanpodila:
closing as stale

Issue #490, opened 2020-05-10
ListChange is not informative for ObservableList changes


darkstarx:
Let's take the method `remove` and setter `length` of the `ObservableList` class:
```
    final list = ObservableList<int>.of(<int>[null, 2, 3]);
    list.observe((ListChange<int> listChange) {
        // (1): index == 0, added == null (!), removed == null, newValue == null, oldValue == null
        //     list: [null, 2, 3, null, null]
        // (2): index == 0, added == null, removed == null (!), newValue == null, oldValue == null
        //     list: [2, 3, null, null]
    });
    list.length = 5;      // (1) append 2 elements with value null
    list.remove(null);  // (2) remove first element with value null
```
Both of them leads to observe identical wrong `ListChange` values, this is bug.

So, fix setter `length`:
```
  @override
  set length(int value) {
    /// There is no need to enforceWritePolicy since we are conditionally wrapping in an Action.
    _context.conditionallyRunInAction(() {
      if (value < _list.length) {
        final removed = _list.sublist(value);
        _list.length = value;
        _notifyListUpdate(value, null, removed);
      } else if (value > _list.length) {
        final index = _list.length;
        _list.length = value;
        _notifyListUpdate(index, _list.sublist(value), null);
      }
    }, _atom);
  }
```
Note: do not notify if the length not changes.

And fix `remove` method:
```
  @override
  bool remove(Object element) {
    var didRemove = false;

    _context.conditionallyRunInAction(() {
      final index = _list.indexOf(element);
      didRemove = _list.remove(element);

      if (didRemove) {
        _notifyListUpdate(index, null, [element]);  // Here show the count of deleted values even if element is null!
      }
    }, _atom);

    return didRemove;
  }
```
Note: there is the same bug in the `removeAt` method: `_notifyListUpdate(index, null, value == null ? null : [value]);`! Why you hides information about the number of removed items? This is bug!
Note: `removeLast` contains the same bug.

And what about `removeWhere` method?
```
      final removedItems = _list.where(test).toList(growable: false);
      _list.removeWhere(test);
      _notifyListUpdate(0, null, removedItems);
```
Are you seriously? How can I define the indexes of removed items? It does matter for flutter AnimatedList for instance.

And what about `replaceRange` method?
Why you send me `ListChange` with `OperationType.add` instead of `OperationType.update`? Were you lazy to extend `_notifyListUpdate` arguments list with optional `OperationType type`?

[...]

PR #496, opened 2020-05-12
fix and improve ListChange details for ObservableList (#490)

darkstarx:


mobx/test/observable_list_test.dart
```diff
@@ -98,20 +99,72 @@ void main() {
...
+    test('observe set length works', () {
+      final list = ObservableList.of([0]);
+
+      var changesCount = 0;
+      var index = -1;
+      var addedValues = <int>[];
+      var removedValues = <int>[];
+
+      list.observe((change) {
+          changesCount += change.rangeChanges?.length ?? 0;
```
pavanpodila, line 111: Can you simplify these tests for all the newly added tests. I think there is excessive conditional logic inside the tests. We can instead make explicit checks for the **before - after** changes, rather than applying conditional logic. That will improve the readability as well maintainability.

mobx/test/observable_list_test.dart
```diff
@@ -98,20 +99,72 @@ void main() {
...
+              index = change.elementChanges.keys.first;
+              final firstChange = change.elementChanges[index];
+              addedValues = [OperationType.add, OperationType.update].contains(firstChange.type)
+                ? [firstChange.newValue]
+                : null;
+              removedValues = [OperationType.remove, OperationType.update].contains(firstChange.type)
+                ? [firstChange.oldValue]
+                : null;
+            }
+          }
```
pavanpodila, line 163: This is one of the example of the excessive conditional logic inside tests. The befor - after checks should be simpler IMHO. Let's have an explicitly created before / after states for comparison

mobx/test/observable_list_test.dart
```diff
@@ -120,18 +173,37 @@ void main() {
...
```
pavanpodila, line 202: same as before

mobx/test/observable_list_test.dart
```diff
@@ -140,84 +212,249 @@ void main() {
...
+        });
```
pavanpodila, line 267: here too and below as well for the remaining newly added tests

mobx/lib/src/api/observable_collections/observable_list.dart
```diff
@@ -60,8 +60,15 @@ class ObservableList<T>
...
-      _notifyListUpdate(0, null, null);
+      if (value < _list.length) {
+        final removed = _list.sublist(value);
+        _list.length = value;
+        _notifyRangeUpdate(value, null, removed);
+      } else if (value > _list.length) {
+        final index = _list.length;
+        _list.length = value;
+        _notifyRangeUpdate(index, _list.sublist(index), null);
+      }
```
pavanpodila, line 71: This was a big miss since the usage was very less. Thanks for pitching in and fixing this part 👍

mobx/lib/src/api/observable_collections/observable_list.dart
```diff
@@ -259,69 +271,122 @@ class ObservableList<T>
...
   }
 
   @override
   void retainWhere(bool Function(T element) test) {
     _context.conditionallyRunInAction(() {
-      final removedItems = _list.where((_) => !test(_)).toList(growable: false);
-
-      _list.retainWhere(test);
-      _notifyListUpdate(0, null, removedItems);
+      final removedElements = <int, ElementChange>{};
```
pavanpodila, line 313: Rather than storing the `removedElements` as a Map, its probably better to store the `index` in the `ElementChange` itself as a field? That way the element-changes would be a list and easier to iterate ? Curious what you think.

pavanpodila: On second thought, using a Map keeps the `ElementChange` agnostic of the `index` itself and as a consumer you would iterate on the Map `keys` instead ?

darkstarx: Thanks!
Actually using the Map ensures that change of the element with specific index will be unique over a single ListChange object. It must be unique, and the interface of ListChange class guaranties this. Probably there was no other serious reasons for me to move the `index` field outside of the ElementChange class.
The same idea applies to the RangeChange class except that there are no checks of ranges overlaying for now.
But you mentioned the sorting, and now I think you were right. The order does much more matter than uniqueness in interface and makes consumers more efficient.

Regarding the unit tests simplifying.. I'm not sure I correctly understand what you mean. Are you talking about using of checks (`expect`) inside of consumer closure body? If so, sure I'll fix these issues.

darkstarx:
I leaved conditions in a few tests in favor of flexibility, in some cases it is allowed in implementation to store the changes either in RegionChange or in ElementChange (add, remove, etc.).
Check the fixes please.

mobx/lib/src/api/observable_collections/observable_list.dart
pavanpodila: Ya the `Map` makes more sense now.

Looks like the coverage has dropped. Can you bring it back to 99%?

pavanpodila: I think unique-ness will be guaranteed as in a single pass of the observe, you will not see simultaneous changes on the same index, assuming we are running this on a single thread!

In that case, have the index inside the ElementChange would be easier to handle. The unit tests looks more readable now, just that the coverage needs to be improved.

darkstarx: It seems that codecov complains only on 2 lines with `assert(index != null)`. I can't realize how to cover these expressions by tests. Could you help me please? How do you usually do this?

darkstarx: Ok, I figured it out. Check this out please.

pavanpodila:
@allContributors add @darkstarx for code,test

pavanpodila:
@allcontributors add @darkstarx for code, test

PR #513, opened 2020-06-04
Update delay in docs

### LINKED ISSUE #512: Correct documentation on reaction delay
Currently the dart docs for `reaction` state:
```
You can also pass in an optional [name], a debouncing [delay] in milliseconds.
``` 
As noted in #101 this delay is actually throttle, not a debounce. To understand the difference, play with this http://demo.nimius.net/debounce_throttle/

I suggest
```
You can also pass in an optional [name], a throttling [delay] in milliseconds.
```

MisterJimson:
Fixes #512

### DIFF
--- a/mobx/lib/src/api/reaction.dart
+++ b/mobx/lib/src/api/reaction.dart
@@ -6,7 +6,7 @@ import 'package:mobx/src/core.dart';
 ///
 /// Optional configuration:
 /// * [name]: debug name for this reaction
-/// * [delay]: debouncing delay in milliseconds
+/// * [delay]: throttling delay in milliseconds
 ///
 /// ```
 /// var x = Observable(10);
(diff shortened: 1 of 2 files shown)


mobx/lib/src/api/reaction.dart
```diff
@@ -6,7 +6,7 @@ import 'package:mobx/src/core.dart';
 ///
 /// Optional configuration:
 /// * [name]: debug name for this reaction
-/// * [delay]: debouncing delay in milliseconds
+/// * [delay]: throttling delay in milliseconds
```
pavanpodila, line 9: There are other reactions like autorun that have the same kind of docs. Would be good to sweep through other reaction APIs and replace debouncing with throttling

MisterJimson: I updated autorun, I will search for others and report back.

MisterJimson:
I updated the only other place I could find "debounce" in the repo.

pavanpodila:
@allContributors add @MisterJimson for doc

PR #522, opened 2020-06-15
add setUsername, setEmail and setPassword action for form examples

xinhaiwang:


mobx_examples/lib/form/form_store.dart
```diff
@@ -45,6 +45,21 @@ abstract class _FormStore with Store {
     ];
   }
 
+  @action
```
pavanpodila, line 48: Why do we need to introduce these? The default setters of a Store have the mutations wrapped in an action

PR #539, opened 2020-07-23
Fixed self reference while passing parent to child

JohnPinto:
Hi, 

I found an error in the documentation. I was trying to replicate the exact code shown, but I was getting an error because I wasn't able to self reference the parent object as an argument to the child. To fix this I've initialized the child inside the parent's constructor, in that scope I was able to call the self reference and carry on coding. It took me a while to get this right as I'm a beginner in Mobx, so I think other beginners might have some trouble with that particular example as well.

Best regards, João Pinto.


pavanpodila:
@all_contributors add @JohnPinto for doc

PR #554, opened 2020-08-20
docs: Api Docs add optional Arguments (Reactions)

hertleinj:
#249 

Feel Free to criticize. Descriptions may not be accurate enough or plain false.
Added the Docs Description to gain a deeper understanding, this by itself was successfull so feel free to discard if insufficient.


pavanpodila:
Closing it for now. We need to revisit these docs and I'll be doing that next month, once I get freed up.

PR #572, opened 2020-10-01
Fixes #529

### LINKED ISSUE #529: MobX Todos example has an unused observable
I was following along with the Todos example and noticed that the tutorial states the following for the AddTodo widget:
_Similarly, here is the AddTodo widget that observes the list.currentDescription and also fires the list.changeDescription action._
https://mobx.netlify.app/examples/todos

The AddTodo widget has no Observer widget. The action is never used. Additionally the currentDescription observable in the TodoList store is never read and only directly written to.

My question being: why is the observable currentDescription not written to via an `@action` instead of directly. The tutorial shows the `changeDescription` method existing in the store but it doesn't exist in the latest version on GitHub.

It might be good to state the purpose of `currentDescription` or remove it to prevent confusion.
My guess for `currentDescription` existing is that potentially you can leave the todo page and come back still having your last entered text in there. And also you could use the current user input to, for example, check if a todo title already exists while the user is typing.

But in the example it serves no purpose.

The AddTodo widget uses the TextController instead to clear the TextField and the TextField's onSubmitted value to pass the description when creating a new Todo.

Maybe I'm missing something, but it would be nice if someone can clarify.

pr-1:
Tasks Done - 
- Removed unused observable currentDescription from TodoList.
- Removed onChange callback from TextField in AddTodo Widget.
- Changed TextInputAction to TextInputAction.done
- Updated tests accordingly. All test cases passed.

### DIFF
--- a/mobx_examples/lib/main.reflectable.dart
+++ b/mobx_examples/lib/main.reflectable.dart
@@ -29,8 +29,9 @@ final _data = <r.Reflectable, r.ReflectorData>{
             7,
             0,
             const prefix0.JsonSerializable(),
-            const <int>[5],
+            const <int>[4],
             const <int>[
+              5,
               6,
               7,
               8,
@@ -50,10 +51,7 @@ final _data = <r.Reflectable, r.ReflectorData>{
               22,
               23,
               24,
-              25,
-              26,
-              27,
-              28
+              25
             ],
             const <int>[],
             1,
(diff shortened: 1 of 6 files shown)


pr-1:
@pavanpodila  Can you please review this. I don't know much about Codecov. If you can tell me is something wrong with my code ? Because in the Github Actions tab it says all checks have successfully passed.

MaikuB (approved):
LGTM.

@pavanpodila  I think there's an issue with duplicate codecov checks. There's one that says it's failed for `flutter_mobx` but then one that passes for `flutter_mobx`. It's at 100% code coverage as well. Not across how that is setup and may need looking into but I think we should be able to ignore it to merge this in.

pavanpodila:
@all-contributors add @pr-1 for example, code

PR #580, opened 2020-10-09
Formatting and fetch plugin versions

leonardocustodio:
Hello,

I've added the section "Format your code" to the documentation. As I've seen that the library score went down in pub.dev in the last update of codegen. My IDE configs are not the default ones.

Also since the plugin versions were very outdated in the docs. Meaning probably no one remembers to update there. I've added the auto-fetch so the docs always stays with the latest version on it.

I've formatted everything that was wrongly formatted including codegen but I'm not sure if there should be a new release just because of formatting so I did not change the version. Leaving it for you guys to decide.

Regards,

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -418,6 +418,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "leonardocustodio",
+      "name": "Leonardo Custodio",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/5619696?v=4",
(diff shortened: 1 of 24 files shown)


pavanpodila (approved):
LGTM

pavanpodila:
@leonardocustodio can you resolve these conflicts ?

pavanpodila:
BTW, @leonardocustodio the changes to fetching the package version as a webpack plugin is brilliant. Thanks! for that 🥇

PR #603, opened 2020-11-26
Null safety

pavanpodila:


acoutts:
Hi- is there any update on this PR?

acoutts:
Had a dig into it now. Seems like we need to use `analyze 1.0.0` because it can tell us whether or not something is nullable:

```
PropertyAccessorElement.type.nullabilitySuffix
```

This gives us the `NullabilitySuffix` enum:
```dart
/// Suffix indicating the nullability of a type.
///
/// This enum describes whether a `?` or `*` would be used at the end of the
/// canonical representation of a type.  It's subtly different the notions of
/// "nullable", "non-nullable", "potentially nullable", and "potentially
/// non-nullable" defined by the spec.  For example, the type `Null` is
/// nullable, even though it lacks a trailing `?`.
enum NullabilitySuffix {
  /// An indication that the canonical representation of the type under
  /// consideration ends with `?`.  Types having this nullability suffix should
  /// be interpreted as being unioned with the Null type.
  question,

  /// An indication that the canonical representation of the type under
  /// consideration ends with `*`.  Types having this nullability suffix are
  /// called "legacy types"; it has not yet been determined whether they should
  /// be unioned with the Null type.
  star,

  /// An indication that the canonical representation of the type under
  /// consideration does not end with either `?` or `*`.
  none
}
```

I migrated my app today and found that basically make it work by making sure the generated nullable computed values had a `?` on the end of their type, and matched the same for actions which set a nullable type.

But sadly there's a number of packages that we depend on but don't have null safe versions ready yet:
```
Package Name     Current     Upgradable  Resolvable  Latest               

direct dependencies:
analyzer         ✗0.41.2     ✗0.41.2     -           ✓1.0.0               
build            ✗1.6.2      ✗1.6.2      -           ✗1.6.3               
build_resolvers  ✗1.5.3      ✗1.5.3      -           ✗1.5.4               
mobx             ✗1.2.1+4    ✗1.2.1+4    -           ✓2.0.0-nullsafety.2  
source_gen       ✗0.9.10+2   ✗0.9.10+2   -           ✗0.9.10+2            

dev_dependencies:
build_runner     ✗1.11.1     ✗1.11.1     -           ✗1.11.3              
build_test       ✗0.10.12+1  ✗0.10.12+1  -           ✗1.3.7               
coverage         ✗0.13.11    ✗0.13.11    -           ✗0.15.2              
logging          ✗0.11.4     ✗0.11.4     -           ✓1.0.0               
mockito          ✗4.1.4      ✗4.1.4      -           ✓5.0.0-nullsafety.7  
No resolution was found. Try running `dart pub upgrade --null-safety --dry-run` to explore why.
```

I was able to do some testing by forcing analyzer to 1.0.0, but we probably don't want to force dependency overrides:
```yaml
dependency_overrides:
  analyzer: ^1.0.0
```


For anyone who wants to use mobx with null safety now, I suggest manually creating all of your observables/actions/computed values, and you should be ok to use it today. Code generation will come at a later date.

aristeu13:
Has there been any update on the dependencies?

pavanpodila:
Still can't migrate mobx_codegen because the dependencies are not null-safe. This prevents the upgrade to sdk `2.12.0` :-(

deadsoul44:
Built value also has code gen but it is able to generate null-safe code. As far as I understand, it is possible to generate null-safe code with non-null-safe code and build packages.

NarHakobyan:
@pavanpodila do you have any plans to release mobx_codegen null safety version?

rserro:
Is it possible to add a prerelease to https://pub.dev/packages/mobx_codegen like https://pub.dev/packages/flutter_mobx and https://pub.dev/packages/mobx?

JEuler:
@pavanpodila I have checked today, https://pub.dev/packages/source_gen is now null safety. Are there any other dependencies in `codegen` needed to be upgraded? Do you need any help?

pavanpodila:
> @pavanpodila I have checked today, https://pub.dev/packages/source_gen is now null safety. Are there any other dependencies in `codegen` needed to be upgraded? Do you need any help?

Thank for the offer @JEuler, I am already in the midst of upgrading the deps and the code as well. Converted the code, but still have a few tests that need to be fixed. Hope to have a release in a few days after some testing.

JEuler:
Cool! Thank you very much! Feel free to ask for any help :)

PR #624, opened 2021-02-07
Fix testimonials image

leonardocustodio:
This fixes two images on testimonials that were broken.

The reason why they were broken is because both user have changed their profile photo on Twitter.

The images should be fetched instead of hardcoding the link.
If anyone wants to take up the challenge, feel free ;)


pavanpodila (approved):

pavanpodila:
@all-contributors add @leonardocustodio for docs

Issue #220, opened 2019-07-10
Support for read-only observables


eltonomicon:
I'd like to define stores where observables are only mutated by actions and cannot be mutated externally of the store. This currently is not possible because the code generator defines get/set properties for observables.  It would be particularly useful to define read-only observable ObservableList<T> properties since you usually never call the setter once it's initialized.

I know one option is to make private observables then provide read-only computed properties for the observables, but a) that seems unnecessarily verbose, and b) it's not really computed, they're just simple accessors. 

For comparison, here's a simple example of how it currently works.

```
# Current approach
class _TestStore with Store {
    @observable
    int value;
}

# Current generated code
mixin _$TestStore on _TestStore, Store {
 final _$valueAtom = Atom(name: '_TestStore.value');

  @override
  int get value {
    _$valueAtom.context.enforceReadPolicy(_$valueAtom);
    _$valueAtom.reportObserved();
    return super.value;
  }

  @override
  set value(int value) {
    _$valueAtom.context.conditionallyRunInAction(() {
      super.value = value;
      _$valueAtom.reportChanged();
    }, _$valueAtom, name: '${_$valueAtom.name}_set');
  }
}
```

Here's one suggestion of an alternate approach.

```
# Alternate approach
class _TestStore with Store {
    int _value;

    @observable(setter: '_setValue')
    int get value
        => _value;

    void _setValue(int value)
        => _value = value;
}

# Alternate generated code
mixin _$TestStore on _TestStore, Store {
 final _$valueAtom = Atom(name: '_TestStore.value');

  @override
  int get value {
    _$valueAtom.context.enforceReadPolicy(_$valueAtom);
    _$valueAtom.reportObserved();
    return super.value;
  }

  @override
  void _setValue(int value) {
    # Same code that was originally in the generated setter
    _$valueAtom.context.conditionallyRunInAction(() {
      super.value = value;
      _$valueAtom.reportChanged();
    }, _$valueAtom, name: '${_$valueAtom.name}_set');
  }
}
```

I think the ObservableList<T> case as just a read-only property with the default @observable annotation should be sufficient, in that case just generate the getter but not the setter.

pavanpodila:
Are you suggesting something like a `@readonly` annotation where a setter won't be generated:

```dart
@readonly
@observable 
int value;
```

eltonomicon:
@pavanpodila I don't think so. If no code at all is generated for a setter then how could `value` be mutated such that this code would run and trigger a change?

```
[...]

PR #630, opened 2021-03-02
Adds support for @readonly annotation

Ascenio:
As discussed in #220 
Makes possible the following code:
```dart
class CounterStore = _CounterStoreBase with _$CounterStore;

abstract class _CounterStoreBase with Store {
  @readonly
  int _counter = 0;

  @action
  void increment() {
    _counter++;
  }
}
```
Such that:
```dart
// This works
final store = CounterStore();
final count = store.counter; // 0

// And this doesn't
final store = CounterStore();
store.counter = -1;
```


mobx_codegen/lib/src/template/observable.dart
```diff
@@ -6,13 +6,24 @@ class ObservableTemplate {
...
+    return name;
+  }
 
   @override
   String toString() => """
   final $atomName = Atom(name: '${storeTemplate.parentTypeName}.$name');
 
   @override
-  $type get $name {
+  $type get $_getterName {
```
pavanpodila, line 26: If its readonly, you need to throw an error in the setter. Don't see that getting handled here

pavanpodila (changes requested):

mobx_codegen/lib/src/template/observable.dart
Ascenio: Actualy it's readonly from the perspective of the client of the store, but it must be modifiable by the store. The setter must stay as is, otherwise it wouldn't be possible to change the value from the variable. The way I did it was to keep the setter private.

Ascenio: Because the variable must be declared as private (e.g.: `@readonly int _counter`) there isn't any setter publicly available.

pavanpodila: So there is still the responsibility of declaring the observable as private at the Store level. That's a hidden assumption and needs to be checked and verified in the codegen too. I am still debating if this is necessary or can be managed in user code itself.

Ascenio: That's true. It would be nice to see other people's opinion about this. For now I can't think of a way to make it such that declaring as private isn't required.

Ascenio: At least if you declared as public should still work. Maybe we can throw a error during generation to inform the error? The same way as actions are not allowed to be used on setters.

pavanpodila: Yes we can do those checks but also ensure the user is not adding their own public setter

Ascenio: Didn't thought about that, good idea. However I'm not really used to codegen, so may take a little while xd.

Ascenio: I managed to disallow public variables, but how can I avoid the setters? Should I search for setters with similar name to the variable?

pavanpodila: yes there should be a way to check the setter on the same name

Ascenio: Done. My solution is basicly caching the public setters found and then checking if there's any that matches the name of a readonly observable. Not sure if that's the right way, but it works.

pavanpodila (changes requested):
all looks good, can you see why the code coverage is not improving even after you add the tests.

Ascenio:
> all looks good, can you see why the code coverage is not improving even after you add the tests.

It's weird because I there is a test for the extension but the coverage doesn't agree with me.

Ascenio:
@pavanpodila could you tell me what is your setup for tests? I'm getting some weird 
`FormatException (FormatException: No ':' on line (at offset 1))` from the `generate` function which makes all tests from _invalid_annotations_test.dart_ and _mobx_codegen_test.dart > generator_ fail

Ascenio:
If you take a look at [codecov for mobx_codegen]( mobx_codegen/lib) it's already at 38.32% coverage, which makes me think that it was already with low coverage when I made a fork.

mobx_codegen/lib/src/errors.dart
```diff
@@ -152,6 +156,16 @@ class InvalidObservableAnnotations extends PropertyErrors {
...
 }
 
+class InvalidReadOnlyAnnotations extends PropertyErrors {
+  @override
+  // ignore: overridden_fields
+  Pluralize propertyPlural = _fieldPluralizer;
+
+  @override
+  String get message =>
+      'Remove @readonly annotation from $property $propertyList. They only apply to private fields.';
```
pavanpodila, line 166: Please check this message. Is private ==> public ?

pavanpodila (changes requested):
Codecov is reporting that some lines are not covered by tests. Can we address that ?

mobx_codegen/lib/src/errors.dart
Ascenio: So you want to change the message? No problem. But what's the new one? 'Is private ==> public'?

pavanpodila: No I mean, is this for private fields or public? Perhaps make it more clear

Ascenio: Oh I see. The annotation @readonly should be used only on private fields to prevent public setters.

Ascenio: I used this message because it was similar to others.

Ascenio:
@pavanpodila ok so it seems like I made progress. I found that the coverage was low because the tests were failing because of some problem with build_test dependency. After I upgraded its version and related packages the tests are now passing and coverage is back at 100%. However, because my change affect both mobx and mobx_codegen at the same time, and I analyzer is screaming because I can't make mobx_codegen depend on the mobx from relative path, what should I do?

pavanpodila:
Ah ok! Let's push out the changes for mobx first. I can publish a pre release, which you can use in codegen. I suggest splitting the MobX only changes into a separate PR

Ascenio:
So should I close this one and open two anothers, one for each package? Also, could you make another branch or should I make PRs to master?

Ascenio:
Sure. I'm going to make the MobX related PR soon.

Ascenio:
Closing this one as it has been replaced by #647 and #649. Both are the same implementation as this one, but are based on null_safety branch.

PR #631, opened 2021-03-05
Support NNBD (Null safety) for generated code of mobx_codegen

fzyzcjy:
Hi thanks for the lib! I try to update mobx_codegen, such that the code that it *generates* will support NNBD.

Related: https://github.com/mobxjs/mobx.dart/issues/598

Also related: https://github.com/mobxjs/mobx.dart/issues/592

An example:

```
class Cls<T> extends _Cls<T> with _$Cls {
  Cls();
}

abstract class _Cls<T> with Store {
  @observable
  List<T>? genericArr;

  @observable
  List<async.Future> futureList;

  @observable
  List<async.Future<double?>>? moreList;

  // -------------------------------------------------------

  _Cls({
    this.genericArr,
    this.futureList,
    this.moreList,
  });
}
```

yields:

```
mixin _$Cls<T> on _Cls<T>, Store {
  final _$genericArrAtom = Atom(name: '_Cls.genericArr');

  @override
  List<T>? get genericArr {
    _$genericArrAtom.reportRead();
    return super.genericArr;
  }

  @override
  set genericArr(List<T>? value) {
    _$genericArrAtom.reportWrite(value, super.genericArr, () {
      super.genericArr = value;
    });
  }

  final _$futureListAtom = Atom(name: '_Cls.futureList');

  @override
  List<Future<dynamic>> get futureList {
    _$futureListAtom.reportRead();
    return super.futureList;
  }

  @override
  set futureList(List<Future<dynamic>> value) {
    _$futureListAtom.reportWrite(value, super.futureList, () {
      super.futureList = value;
    });
  }

  final _$moreListAtom = Atom(name: '_Cls.moreList');

  @override
  List<Future<double?>>? get moreList {
    _$moreListAtom.reportRead();
    return super.moreList;
  }

  @override
  set moreList(List<Future<double?>>? value) {
    _$moreListAtom.reportWrite(value, super.moreList, () {
      super.moreList = value;
    });
  }

  @override
  String toString() {
    return '''
genericArr: ${genericArr},
futureList: ${futureList},
moreList: ${moreList}
    ''';
  }
}
```

PR #634, opened 2021-03-07
Support NNBD (Null safety) for generated code of mobx_codegen (new)

fzyzcjy:
Same code as https://github.com/mobxjs/mobx.dart/pull/631

### DIFF
--- a/mobx_codegen/lib/src/template/computed.dart
+++ b/mobx_codegen/lib/src/template/computed.dart
@@ -10,7 +10,7 @@ class ComputedTemplate {
   @override
   // ignore: prefer_single_quotes
   String toString() => """
-  Computed<$type> $computedName;
+  Computed<$type>? $computedName;
 
   @override
   $type get $name => ($computedName ??= Computed<$type>(() => super.$name, name: '${storeTemplate.parentTypeName}.$name')).value;""";
--- a/mobx_codegen/lib/src/template/params.dart
+++ b/mobx_codegen/lib/src/template/params.dart
@@ -3,17 +3,16 @@ class ParamTemplate {
   String type;
   String defaultValue;
   bool hasRequiredAnnotation = false;
+  bool hasRequiredKeyword = false;
 
   String get asArgument => name;
 
   NamedArgTemplate get asNamedArgument => NamedArgTemplate()..name = name;
 
-  String get metadata => hasRequiredAnnotation ? '@required ' : '';
+  String get metadata => (hasRequiredAnnotation ? '@required ' : '') + (hasRequiredKeyword ? 'required ' : '');
 
   @override
-  String toString() => defaultValue == null
-      ? '$metadata$type $name'
-      : '$type $name = $defaultValue';
+  String toString() => defaultValue == null ? '$metadata$type $name' : '$type $name = $defaultValue';
 }
 
 class TypeParamTemplate {
--- a/mobx_codegen/lib/src/template/params.dart
+++ b/mobx_codegen/lib/src/template/params.dart
@@ -2,14 +2,13 @@ class ParamTemplate {
   String name;
   String type;
   String defaultValue;
-  bool hasRequiredAnnotation = false;
   bool hasRequiredKeyword = false;
 
   String get asArgument => name;
 
   NamedArgTemplate get asNamedArgument => NamedArgTemplate()..name = name;
 
-  String get metadata => (hasRequiredAnnotation ? '@required ' : '') + (hasRequiredKeyword ? 'required ' : '');
+  String get metadata => hasRequiredKeyword ? 'required ' : '';
 
   @override
   String toString() => defaultValue == null ? '$metadata$type $name' : '$type $name = $defaultValue';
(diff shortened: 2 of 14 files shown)


mobx_codegen/lib/src/template/computed.dart
```diff
@@ -10,7 +10,7 @@ class ComputedTemplate {
   @override
   // ignore: prefer_single_quotes
   String toString() => """
-  Computed<$type> $computedName;
+  Computed<$type>? $computedName;
```
pavanpodila, line 13: Why are you forcing nullable here for the property ? The Computed is always non-nullable, instead the value it produces can potentially be nullable depending on how the user sets it up. Please remove the `?' here

mobx_codegen/lib/src/template/params.dart
```diff
@@ -3,17 +3,16 @@ class ParamTemplate {
...
   String defaultValue;
   bool hasRequiredAnnotation = false;
+  bool hasRequiredKeyword = false;
 
   String get asArgument => name;
 
   NamedArgTemplate get asNamedArgument => NamedArgTemplate()..name = name;
 
-  String get metadata => hasRequiredAnnotation ? '@required ' : '';
+  String get metadata => (hasRequiredAnnotation ? '@required ' : '') + (hasRequiredKeyword ? 'required ' : '');
```
pavanpodila, line 12: I think the annotation is no longer valid in this version of Dart, let's just use the keyword check instead

pavanpodila (changes requested):
much cleaner...just a few more changes

mobx_codegen/lib/src/template/computed.dart
fzyzcjy: But we need to support `$computedName ??= Computed()`, so it should be nullable

mobx_codegen/lib/src/template/params.dart
fzyzcjy: ok

mobx_codegen/lib/src/template/computed.dart
pavanpodila: That is for the name of the computed property, not the Computed type

pavanpodila (changes requested):

mobx_codegen/lib/src/template/computed.dart
fzyzcjy: I am confused... if remove "?" then error.

fzyzcjy: Normally (notice the type is `int?`here, but often it is things like `int` in other places:

image


If no "?"

image

If use late:

image

fzyzcjy:
@pavanpodila See the comments above

mobx_codegen/lib/src/template/computed.dart
pavanpodila: ya I see the issue now. Computed is created lazily on the first access, so I think late is the right approach. Not sure why greyed out text. Can you replace with `late` for now?

pavanpodila (changes requested):

mobx_codegen/lib/src/template/computed.dart
pavanpodila: the initialization of the computed variable can be move to the declaration once we make it lazy. In fact we can make it `final lazy`

pavanpodila (changes requested):

fzyzcjy:
@pavanpodila Hmm when I do this, error:

image

pavanpodila:
Aah looks like only this reference is allowed with late variables. Alright let's go back to nullable 👍

fzyzcjy:
@pavanpodila So when can we release the new version ;)

pavanpodila (approved):

pavanpodila:
> @pavanpodila So when can we release the new version ;)

I updated the sdk version to `2.12.0` and looks like I had a ton of errors on null-safety ;-). Will try to fix this today and hopefully a release in a day or so 🤔

pavanpodila:
Sadly we cannot upgrade to 2.12.0, just yet because there are some dependencies that have not been migrated, esp. the build and source_gen packages. We will have to suppress the legacy import warning and make some changes to see if it works. 

Try changing the sdk version to `2.12.0` and see how far you can go ?

fzyzcjy:
Let me have a look. Actually we may use 2.12 and solve the errors, while some dependencies are legacy?

fzyzcjy:
See this: https://github.com/fzyzcjy/mobx.dart/tree/new_null_safety_2

Test can be run by `dart --no-sound-null-safety test `

pavanpodila:
I'll give it a try later today or tomorrow. Have a bunch of client deliverables this week.

hacker1024:
There are dependency conflicts with the latest version of `build_runner`.

```
Because every version of mobx_codegen from git depends on build_resolvers ^1.5.4 and build_runner >=1.12.0 depends on build_resolvers ^2.0.0, mobx_codegen from git is incompatible with build_runner >=1.12.0.
So, because app depends on both build_runner ^1.12.1 and mobx_codegen from git, version solving failed.
pub get failed (1; So, because app depends on both build_runner ^1.12.1 and mobx_codegen from git, version solving failed.)
```

PR #654, opened 2021-04-05
Adds @readonly to the docs

Ascenio:
Also updated the summary to simplify searching for `@foo_bar`. Let me know if you don't want this behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,9 @@ Language: [English](README.md) | [Português](translation/pt-BR/README.md) | [Ch
 - **[Introduction](#introduction)**
 - **[Core Concepts](#core-concepts)**
   - [Observables](#observables)
-  - [Computed Observables](#computed-observables)
+    - [@observable](#observables)
+    - [@readonly](#readonly)
+    - [@computed](#computed-observables)
   - [Actions](#actions)
   - [Reactions](#reactions)
 - **[Contributing](#contributing)**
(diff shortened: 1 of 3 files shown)


pavanpodila:
We need it on the docs for website, not just readme

Ascenio:
Under which section should I add it? Maybe under `Core Concepts` with the same body?

pavanpodila:
Put under api

Ascenio:
But then should I modify the body of the README to be simpler? I can only think of a before and after comparison to make it shorter.

pavanpodila:
Yes. Readme cannot be long. Keep the details, examples in the Api page

pavanpodila (changes requested):
I think we should shorten the description of read-only in readme. Also move it further down in order

Ascenio:
What about the order? Should it be bellow computed but still under observables' section?

README.md
```diff
@@ -27,7 +27,9 @@ Language: [English](README.md) | [Português](translation/pt-BR/README.md) | [Ch
 - **[Introduction](#introduction)**
 - **[Core Concepts](#core-concepts)**
   - [Observables](#observables)
-  - [Computed Observables](#computed-observables)
+    - [@observable](#observables)
+    - [@readonly](#readonly)
```
pavanpodila, line 31: I suggest removing from readme as this detail is not needed at this level

README.md
```diff
@@ -133,6 +135,44 @@ will fade away and you will mostly focus on the code within the braces.
 
 **Note**: Annotations are available via the **[mobx_codegen](https://github.com/mobxjs/mobx.dart/tree/master/mobx_codegen)** package.
 
+### Readonly
```
pavanpodila, line 138: all of this can be move to the `observables.mdx`

pavanpodila (changes requested):

README.md
Ascenio: I have already moved the previous README to `observables.mdx`. Should I just remove the readonly from the README?

pavanpodila: Ya let's remove from readme

Ascenio: Done

Ascenio:
@pavanpodila I think it's weird to leave the reference out of the README because users who don't access the website won't know it exists. Anyway, the references to the site have been added and the ones from the README have been removed already.

pavanpodila:
We can keep 2-3 lines about read only in the main Readme.

Ascenio:
I've simplified both READMEs and left the detailed explanation to the site.

pavanpodila (approved):

PR #669, opened 2021-05-25
Enable release mode spying

s0nerik:
Currently, spying on the [ReactiveContext] in release builds is prohibited. This makes it impossible to log state-management events for remote diagnostics of user problems in production, which is a great use-case for `@action`s, IMO.

This PR removes the `_isDebugMode` check within the `ReactiveContext.isSpyEnabled`, thus allowing spying in release mode if needed.

### DIFF
--- a/mobx/lib/src/core/context.dart
+++ b/mobx/lib/src/core/context.dart
@@ -60,7 +60,7 @@ class ReactiveConfig {
     this.writePolicy = ReactiveWritePolicy.observed,
     this.readPolicy = ReactiveReadPolicy.never,
     this.maxIterations = 100,
-    this.isSpyEnabled = true,
+    this.isSpyEnabled = false,
   });
 
   /// The main or default configuration used by [ReactiveContext]
--- a/mobx/lib/src/core/context.dart
+++ b/mobx/lib/src/core/context.dart
@@ -60,6 +60,7 @@ class ReactiveConfig {
     this.writePolicy = ReactiveWritePolicy.observed,
     this.readPolicy = ReactiveReadPolicy.never,
     this.maxIterations = 100,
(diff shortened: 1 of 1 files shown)


OlehMarchenko95 (approved):
LGTM

mobx/lib/src/core/context.dart
```diff
@@ -60,6 +60,7 @@ class ReactiveConfig {
     this.writePolicy = ReactiveWritePolicy.observed,
     this.readPolicy = ReactiveReadPolicy.never,
     this.maxIterations = 100,
+    this.isSpyEnabled = true,
```
pavanpodila, line 63: By default let's keep it false and have it be explicitly enabled.

pavanpodila (changes requested):

mobx/lib/src/core/context.dart
s0nerik: I've changed this piece as you described, but I'd like to note that the only way to enable spying now is by having a code that changes the `mainContext`'s config at runtime, like this:

```dart
if (needToSpy) {
  mainContext.config = ReactiveConfig.main.clone(isSpyEnabled: true); // or ReactiveConfig(isSpyEnabled: true)
  mainContext.spy(_handleMobxEvent);
}
```

instead of just having something like this:
```dart
if (needToSpy) {
  mainContext.spy(_handleMobxEvent);
}
```

The former looks acceptable to me, but I can't imagine a case where no-spying-by-default policy can be useful. Maybe you have some thoughts on this, @pavanpodila ?

pavanpodila (approved):

pavanpodila:
@all-contributors add @s0nerik for code

Issue #681, opened 2021-06-20
ObservableStream.value throws Exception when data is null


moritz-weber:
First of all: Thank you for your work :)

I stumbled across an error when migrating to null-safety.
The problematic code is in `mobx/lib/src/api/async/observable_stream.dart` (around line 42):

    /// Current value or null if waiting and no initialValue, or null if data is an error.
    @override
    T? get value =>
        _controller.valueType == _ValueType.value ? _controller.data as T : null;

In my case, no error has occurred so `valueType == value` but `data` is null.
Therefore, mobx tries to cast null to T and throws an ugly error.
I could manually check for `waiting` before accessing `value`, but I'm lazy and think this should be fixed here ;)

pavanpodila:
How about a PR?

moritz-weber:
Just created the PR. Let me know if this works for you. I don't contribute code to other projects very often.

windrunner414:
hi, when will a release contain this PR? The latest version was released 3 months ago. I also faced this issue and now I have to use dependency_overrides to let it work

pavanpodila:
Hi, my laptop had to be changed and I'll take a day or two to get it back. Should be published this week.

pavanpodila:
@windrunner414 this is now published: https://pub.dev/packages/mobx

PR #682, opened 2021-06-20
Fix for #681: ObservableStream.value throws Exception when data is null

moritz-weber:


pavanpodila (changes requested):
Can we get a test as well to make sure this is working? You can use the other tests for reference

moritz-weber:
Writing the test is not as easy as I thought.
The existing test `providing no initialValue sets initial status to waiting` covers the case that I want to test.
However, it doesn't throw an exception as the tests are run with Dart 2.10 (from what I understand).

pavanpodila (approved):

pavanpodila:
@moritz-weber Please do a merge from master and refresh this PR. I made a few fixes in master for the analysis errors. That should clear the Github checks.

pavanpodila (approved):

pavanpodila:
@all-contributors add @moritz-weber for code

PR #697, opened 2021-08-02
Bug fix for Computed with nullable values

brianrobles204:
Hello,

While working on my app, I noticed that Computed will sometimes not propagate changes in the following simplified use case:

```dart
final emailComputed = Computed<String?>((){
    final username = usernameObservable.value;
    final user = getUserOrThrow(username);
    return user?.email; // can be null
});
```

In the case where callback initially throws, then later recomputes successfully but returns `null`, the computed thinks that nothing changed. This is because there's no caught exception and the value before and after are both null. So it doesn't propagate changes, and any further derivations still think the computed has a caught error

This pull request fixes the bug and ensures that computed with nullable values will propagate changes after errors correctly. Thanks for your time!

### DIFF
--- a/mobx/lib/src/core/computed.dart
+++ b/mobx/lib/src/core/computed.dart
@@ -138,12 +138,14 @@ class Computed<T> extends Atom implements Derivation, ObservableValue<T> {
 
     final oldValue = _value;
     final wasSuspended = _dependenciesState == DerivationState.notTracking;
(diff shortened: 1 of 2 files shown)


pavanpodila (approved):
Great edge case and fix. Good on tests too!

pavanpodila:
@all-contributors add @brianrobles204 for code and test

PR #708, opened 2021-09-24
Fix: ObservableStream.listen() should also keep observable values updated

brianrobles204:
Hello! 

This PR addresses a TODO in ObservableStream where its observable values (valueType, data, status) are not updated if they are not observed by MobX, even if the underlying stream is emitting events already because `listen()` was called. 
For example:

```dart
final stream = ObservableStream(sourceStream);
final sub = stream.listen(null); // stream starts running

// ...later

print(stream.value) // read outside reaction, still prints null even though stream has emitted values
```

As a side effect, it's now possible to listen to a single-subscription stream *and* observe its values in a reaction, at the same time :) Previously, trying to do both would throw. Note that to comply with the single-subscription stream contract, calling `cancel` on the subscription will also stop further updates to observable values.

Other changes for single-subscription streams:
- InitialValue is also emitted as the first value for subscriptions. This means calling `listen` and observing `value` should both emit the exact same list of values
- Added a `close` method to force close the observable stream. This is only needed in some cases for single-sub streams because if the subscription is paused and never resumed, the parent stream cannot close correctly. See explanation in docs / tests

For broadcast source streams, there should be mostly no difference, other than the observable values should now always be updated (as long as there is at least one mobx observer or stream subscription). 

All existing tests still pass and I added new ones to cover the above behaviors. I needed this fix for an edge case in my app and so far after testing, things work as expected. Let me know if you agree with the above behaviors or if you think there needs to be other changes. Thanks!

### DIFF
--- a/mobx/lib/src/api/async/observable_stream.dart
+++ b/mobx/lib/src/api/async/observable_stream.dart
@@ -300,7 +300,7 @@ class _ObservableStreamController<T> {
             onListen: _listen,
             onPause: _unsubscribe,
             onResume: _listen,
-            onCancel: _unsubscribe,
+            onCancel: _onCancel,
             sync: true,
           );
   }
@@ -309,6 +309,7 @@ class _ObservableStreamController<T> {
   final bool cancelOnError;
   final Stream<T> origStream;
   StreamSubscription<T>? _subscription;
+  StreamSubscription<T>? _broadcastSubscription;
   T? _initialStreamValue;
 
   final ActionController _actions;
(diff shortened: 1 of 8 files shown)


pavanpodila:
@brianrobles204 , there are some Analyze errors, which if fixed can pass the build

pavanpodila (approved):
LGTM, just need to fix the analyzer errors to pass build

pavanpodila:
@brianrobles204, please see if you can add documentation for this behavior in the docs

brianrobles204:
Sure will do, thanks!

brianrobles204:
Hi, I fixed the analyzer issues, they were being caused by additional dart lints I think? 

Also added some documentation on the stream behaviors I outlined. Let me know if it's clear

The code coverage issues seem to be existing coverage gaps? I mostly just edited observable_stream.dart and its test file, and they have 100% coverage. The other changed files were just for the analyzer

pavanpodila:
@all-contributors add @brianrobles204 for code

PR #721, opened 2021-12-21
docs: adds context configuration for spies

Ascenio:
Hello :wave: 

I've noticed that the section about _spying_ in the docs is somewhat incomplete because **mainContext**'s `isSpyEnabled` defaults to false, so if the user bindly copy the samples they won't work.

### DIFF
--- a/docs/docs/api/spy.mdx
+++ b/docs/docs/api/spy.mdx
@@ -23,8 +23,9 @@ Using the `spy()` method of a `ReactiveContext`, you can get notified of all act
 import 'package:mobx/mobx.dart';
 
 void main() {
-
-  // ...
+  mainContext.config = mainContext.config.clone(
+    isSpyEnabled: true,
+  );
 
   mainContext.spy((event) { /* ... */ });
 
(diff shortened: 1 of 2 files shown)


Ascenio:
Hey, @pavanpodila. Could you, please, re-run the CI? #722 has it fixed it! ;D

docs/docs/api/spy.mdx
```diff
@@ -34,6 +35,9 @@ void main() {
 
 In this case, we are setting up a spy right in the top-level `main()` method. This gives us visibility into MobX, right from the get-go.
 
+Notice that before we spy on events, we must firstly make sure we enable
```
pavanpodila, line 38: Minor grammar fix:

**"we must first make sure that spying is enabled."**

pavanpodila (changes requested):

mobx/lib/src/api/observable_collections/observable_map.dart
```diff
@@ -73,7 +73,8 @@ class ObservableMap<K, V>
     _context.enforceReadPolicy(_atom);
 
     _atom.reportObserved();
-    return _map[key as K?];
+    final castedKey = key as K?;
```
pavanpodila, line 76: Please update from master to change these lines. We are adopting a slightly different approach

pavanpodila (changes requested):

PR #723, opened 2021-12-21
Perf/increase performance when spy is disabled

Ascenio:
## What
This PR increases [mobx](https://pub.dev/packages/mobx) performance by avoiding creating Strings and DateTime objects when _spying_ is off.

## Benchmarking code
```dart

part 'main.g.dart';

class CounterStore = _CounterStoreBase with _$CounterStore;

abstract class _CounterStoreBase with Store {
  @observable
  int counter = 0;

  @action
  void increment() {
    counter++;
  }
}

void main(List<String> args) {
  final counter = CounterStore();
  final watch = Stopwatch()..start();
  for (var i = 0; i < 1000000; i++) {
    counter.increment();
  }
  watch.stop();
  print(watch.elapsedMilliseconds);
}
```

## Results

In my laptop, it went down from **~530ms** to shocking **~250ms** :laughing:.

You can also take a look at the bottom-up info taken from [Dart DevTools](https://dart.dev/tools/dart-devtools) bellow. Notice how our bottleneck reduced.

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
</td>
<td>
</td>
</tr>
</table>

### DIFF
--- a/mobx/lib/src/api/observable_collections/observable_map.dart
+++ b/mobx/lib/src/api/observable_collections/observable_map.dart
@@ -73,7 +73,8 @@ class ObservableMap<K, V>
     _context.enforceReadPolicy(_atom);
 
     _atom.reportObserved();
-    return _map[key as K?];
+    final castedKey = key as K?;
+    return _map[castedKey];
   }
 
   @override
--- a/mobx/lib/src/core/atom_extensions.dart
+++ b/mobx/lib/src/core/atom_extensions.dart
@@ -10,7 +10,7 @@ extension AtomSpyReporter on Atom {
     context.spyReport(ObservableValueSpyEvent(this,
         newValue: newValue, oldValue: oldValue, name: name));
 
-    final actionName = context.isSpyEnabled ? '${name}_set' : null;
+    final actionName = context.isSpyEnabled ? '${name}_set' : name;
 
     // ignore: cascade_invocations
     context.conditionallyRunInAction(() {
--- a/mobx/lib/src/core/atom_extensions.dart
+++ b/mobx/lib/src/core/atom_extensions.dart
@@ -10,11 +10,13 @@ extension AtomSpyReporter on Atom {
     context.spyReport(ObservableValueSpyEvent(this,
         newValue: newValue, oldValue: oldValue, name: name));
 
+    final actionName = context.isSpyEnabled ? '${name}_set' : null;
+
     // ignore: cascade_invocations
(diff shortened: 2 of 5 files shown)


Ascenio:
Hey, @pavanpodila. Could you, please, re-run the CI? #722 has it fixed it! ;D

mobx/lib/src/core/atom_extensions.dart
```diff
@@ -10,11 +10,13 @@ extension AtomSpyReporter on Atom {
     context.spyReport(ObservableValueSpyEvent(this,
         newValue: newValue, oldValue: oldValue, name: name));
 
+    final actionName = context.isSpyEnabled ? '${name}_set' : null;
```
pavanpodila, line 13: `actionName` should be set to `name` when spy is disabled

pavanpodila (changes requested):
just a minor change, rest looks good

mobx/lib/src/core/atom_extensions.dart
pavanpodila: Also @Ascenio, there are some analyser errors: https://github.com/mobxjs/mobx.dart/runs/4590667181?check_suite_focus=true

Ascenio: About the analyzer errors, they were are already in mobx's master. PR #722 fixed it.

Ascenio: Do you mean like this?
```suggestion
    final actionName = context.isSpyEnabled ? '${name}_set' : name;
```

pavanpodila: Yes

mobx/lib/src/api/observable_collections/observable_map.dart
```diff
@@ -73,7 +73,8 @@ class ObservableMap<K, V>
     _context.enforceReadPolicy(_atom);
 
     _atom.reportObserved();
-    return _map[key as K?];
+    final castedKey = key as K?;
```
pavanpodila, line 76: Please update from master to remove these lines. We are going with a simpler approach

pavanpodila (changes requested):

mobx/lib/src/api/observable_collections/observable_map.dart
pavanpodila: I've committed a fix for this

PR #726, opened 2021-12-23
Change from `late` to `late final` to make the code a bit stronger

fzyzcjy:
should not break anything imho

### DIFF
--- a/flutter_mobx/lib/src/observer_widget_mixin.dart
+++ b/flutter_mobx/lib/src/observer_widget_mixin.dart
@@ -70,7 +70,7 @@ mixin ObserverElementMixin on ComponentElement {
 
   @override
   Widget build() {
-    late Widget built;
+    late final Widget built;
 
     reaction.track(() {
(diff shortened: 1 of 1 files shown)


flutter_mobx/lib/src/observer_widget_mixin.dart
```diff
@@ -70,7 +70,7 @@ mixin ObserverElementMixin on ComponentElement {
 
   @override
   Widget build() {
-    late Widget built;
+    late final Widget built;
```
pavanpodila, line 73: Can't be final as the `built` widget is created in every build cycle

pavanpodila:
Can' be late final as the variable changes every build cycle

fzyzcjy:
Well imho it is a *local* variable, not a global one. So this can be final.

pavanpodila:
You are right. Missed that part. Point taken

fzyzcjy:
😄

Issue #718, opened 2021-12-14
Allow custom ReactiveContext creation with codegen


SaeedMasoumi:
Is it possible to pass custom context? something like this:
```
@context
ReactiveContext get context=> mycontext();

PR #770, opened 2022-03-07
Feat: Allow use custom context

amondnet:
close https://github.com/mobxjs/mobx.dart/issues/718

```dart
class CustomContextStore = _CustomContextStore with _$CustomContextStore;

final ReactiveContext customContext = ReactiveContext();

abstract class _CustomContextStore with Store {
  @observable
  late String name;

  @override
  ReactiveContext get context => customContext;
}
```

```dart
mixin _$CustomContextStore on _CustomContextStore, Store {
  late final _$nameAtom =
      Atom(name: '_CustomContextStore.name', context: context);

  @override
  String get name {
    _$nameAtom.reportRead();
    return super.name;
  }

  @override
  set name(String value) {
    _$nameAtom.reportWrite(value, super.name, () {
      super.name = value;
    });
  }

  @override
  String toString() {
    return '''
name: ${name}
    ''';
  }
}
```

### DIFF
--- a/mobx/test/store_test.dart
+++ b/mobx/test/store_test.dart
@@ -1,0 +1,12 @@
+import 'package:mobx/mobx.dart';
+import 'package:test/test.dart';
+
+class TestStore with Store {}
+
+void main() {
+  group('Store', () {
+    test('can get context', () {
+      expect(TestStore().context, mainContext) ;
+    });
+  });
+}
--- a/mobx/test/store_test.dart
+++ b/mobx/test/store_test.dart
@@ -3,10 +3,23 @@ import 'package:test/test.dart';
 
 class TestStore with Store {}
 
+final customContext = ReactiveContext();
+
+class CustomStore with Store {
+  @override
+  ReactiveContext get context => customContext;
+}
+
 void main() {
   group('Store', () {
     test('can get context', () {
-      expect(TestStore().context, mainContext) ;
+      final store = TestStore();
+      expect(store.context, mainContext);
+    });
+
+    test('Store with custom context', () {
+      final store = CustomStore();
+      expect(store.context, customContext);
     });
   });
 }
--- a/mobx_codegen/pubspec.yaml
+++ b/mobx_codegen/pubspec.yaml
@@ -12,7 +12,8 @@ dependencies:
   build: ^2.2.1
   build_resolvers: ^2.0.6
   meta: ^1.3.0
-  mobx: ^2.0.5
+  mobx: #^2.0.5
+    path: ../mobx
   path: ^1.8.0
   source_gen: ^1.2.1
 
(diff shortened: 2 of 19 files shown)


mobx_codegen/test/data/valid_import_prefixed_output.dart
```diff
@@ -138,8 +143,8 @@ mixin _$User<T extends io.Process> on UserBase<T>, Store {
     });
   }
 
-  final _$friendWithExplicitTypeArgumentAtom =
-  Atom(name: 'UserBase.friendWithExplicitTypeArgument');
+  late final _$friendWithExplicitTypeArgumentAtom =
```
fzyzcjy, line 146: I am a bit worried about the cost of "late final", since it imposes runtime checks

mobx_codegen/pubspec.yaml
```diff
@@ -12,7 +12,8 @@ dependencies:
   build: ^2.2.1
   build_resolvers: ^2.0.6
   meta: ^1.3.0
-  mobx: ^2.0.5
+  mobx: #^2.0.5
+    path: ../mobx
```
fzyzcjy, line 16: Maybe we should not merge this temporary operation into master?

mobx_codegen/test/data/valid_import_prefixed_output.dart
amondnet: @fzyzcjy Yes I agree.

Another way I've considered is:
```dart

/// A base class to use custom context.
class Foo<ReactiveContext> {
  final ReactiveContext context;

  Foo(this.context);
}

class CustomContextStore = _CustomContextStore with _$CustomContextStore;

final ReactiveContext customContext = ReactiveContext();

abstract class _CustomContextStore extends Foo with Store {
  @observable
  late String name;

  _CustomContextStore(context) : super(context);

}

void main() {
  test('use custom context', () {
    final store = CustomContextStore(customContext);

    expect(store._$nameAtom.context, equals(customContext));
  });
}
```
However, there is other downside to using `extends`. Is there any other good options?

fzyzcjy: Agree. Have not come up with other approaches yet

fzyzcjy (approved):
LGTM

amondnet:
@fzyzcjy I added test for `Store.context`

mobx/test/store_test.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:mobx/mobx.dart';
+import 'package:test/test.dart';
+
+class TestStore with Store {}
+
+void main() {
+  group('Store', () {
+    test('can get context', () {
+      expect(TestStore().context, mainContext) ;
```
fzyzcjy, line 9: Sounds like the code is not formatted?

pavanpodila (changes requested):
Need some more unit tests in places where Codecov has flagged

pavanpodila (approved):

pavanpodila:
@all-contributors add @amondnet for code

PR #778, opened 2022-03-18
Routine PR (2022.03.18)

### LINKED ISSUE #692: Globally disable `toString()` method generation
Hi thanks for this lib! I wonder whether there is a global configuration parameter to turn off the toString generation?

It is anology to https://pub.dev/packages/json_serializable#build-configuration , where it says:

> 
> Besides setting arguments on the associated annotation classes, you can also configure code generation by setting values in build.yaml.
> 
> ```
> targets:
>   $default:
>     builders:
>       json_serializable:
>         options:
>           # Options configure how source code is generated for every
>           # `@JsonSerializable`-annotated class in the package.
>           #
>           # The default value for each is listed.
>           any_map: false
>           checked: false
>           create_factory: true
>           create_to_json: true
>           disallow_unrecognized_keys: false
>           explicit_to_json: false
>           field_rename: none
>           generic_argument_factories: false
>           ignore_unannotated: false
>           include_if_null: true
> ```

### LINKED ISSUE #776: Allow disabling `No observables detected in the build method of Observer`
The `No observables detected in the build method of Observer` warning message is quite helpful sometimes. However, in many other times, it is not helpful: My Observer conditionally uses some observables, and sometimes it just does not use any. 

My current workaround is to make a dummy function and call it to suppress the error:

```
final _dummyAtom = Atom(name: '_dummyAtom');

void mobxDummyReportRead() => _dummyAtom.reportRead();

T withMobxDummyReportRead<T>(T ret) {
  mobxDummyReportRead();
  return ret;
}
```

But that does not seem to be very elegant.

Maybe an option in `Observer` is enough

### LINKED ISSUE #777: Improve `debugFindConstructingStackFrame` to let the message be less confusing to the reader
Currently it is like:

```
flutter: No observables detected in the build method of Observer
flutter: #3      some_method (package:some_file.dart:74:12)
```

But the reader may be a bit confused about what is that line. So maybe we can improve the message.

### LINKED ISSUE #779: The `analyzer: ^3.0.0` in master branch is incompatible with Flutter stable
If we use that, we see:

```
Because test >=1.20.0 depends on test_api 0.4.9 and no versions of test match >1.19.5 <1.20.0, test >1.19.5 requires test_api 0.4.9.
And because test 1.19.5 depends on analyzer >=1.0.0 <3.0.0, test >=1.19.5 requires analyzer >=1.0.0 <3.0.0 or test_api 0.4.9.
And because every version of flutter_test from sdk depends on test_api 0.4.8 and every version of mobx_codegen from path depends on analyzer ^3.0.0, one of flutter_test from sdk or test >=1.19.5 or mobx_codegen from path must be false.
And because vision_utils_flutter_wrapper depends on si_flutter from path which depends on test ^1.19.5, flutter_test from sdk is incompatible with mobx_codegen from path.
So, because vision_utils_flutter_wrapper depends on both flutter_test from sdk and mobx_codegen from path, version solving failed.
```

### LINKED ISSUE #780: If `builder` of `Observer` errors, further error `LateInitializationError: Local 'built' has not been initialized.` will happen in addition to the actual error, reducing developer experience

fzyzcjy:
Closes #777
Closes #776
Closes #692
Closes #779
Closes #780

### DIFF
--- a/flutter_mobx/lib/src/observer_widget_mixin.dart
+++ b/flutter_mobx/lib/src/observer_widget_mixin.dart
@@ -65,7 +65,8 @@ mixin ObserverElementMixin on ComponentElement {
         library: 'flutter_mobx',
         exception: e,
         stack: e is Error ? e.stackTrace : null,
-        context: ErrorDescription('From reaction of ${_widget.getName()} of type $runtimeType.'),
+        context: ErrorDescription(
+            'From reaction of ${_widget.getName()} of type $runtimeType.'),
       ));
     }) as ReactionImpl;
     super.mount(parent, newSlot);
--- a/flutter_mobx/lib/src/observer_widget_mixin.dart
+++ b/flutter_mobx/lib/src/observer_widget_mixin.dart
@@ -5,6 +5,9 @@ import 'package:mobx/mobx.dart';
 // ignore: implementation_imports
 import 'package:mobx/src/core.dart' show ReactionImpl;
 
+/// Whether to warn when there is no observables in the builder function
+bool enableWarnWhenNoObservables = true;
+
 /// Observer observes the observables used in the `build` method and rebuilds
 /// the Widget whenever any of them change. There is no need to do any other
 /// wiring besides simply referencing the required observables.
@@ -115,7 +118,9 @@ mixin ObserverElementMixin on ComponentElement {
       built = super.build();
     });
 
-    if ((_widget.warnWhenNoObservables ?? true) && !reaction.hasObservables) {
+    if (enableWarnWhenNoObservables &&
+        (_widget.warnWhenNoObservables ?? true) &&
+        !reaction.hasObservables) {
       _widget.log(
         'No observables detected in the build method of ${reaction.name}',
       );
--- a/flutter_mobx/lib/src/observer_widget_mixin.dart
+++ b/flutter_mobx/lib/src/observer_widget_mixin.dart
@@ -42,6 +42,10 @@ mixin ObserverWidgetMixin on Widget {
     debugPrint(msg);
   }
 
+  /// Whether to warn when there is no observables in the builder function
+  /// null means true
+  bool? get warnWhenNoObservables => null;
+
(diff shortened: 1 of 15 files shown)


fzyzcjy:
Seems to need a review before this can be merged (after a few weeks)

flutter_mobx/lib/src/observer_widget_mixin.dart
```diff
@@ -105,19 +114,27 @@ mixin ObserverElementMixin on ComponentElement {
...
       _widget.log(
         'No observables detected in the build method of ${reaction.name}',
       );
     }
 
-    return built;
+    // Better than a "LateInitializationError" which confuses the user, see #780
+    if (built == null) {
+      throw Exception(
+          'Error happens when building ${_widget.runtimeType}, but it was captured since disableErrorBoundaries==true');
```
pavanpodila, line 134: the english is not clear. It has to be described in the past tense.

flutter_mobx/test/flutter_mobx_test.dart
```diff
@@ -98,7 +98,8 @@ void main() {
 
     expect(tester.firstWidget(find.byWidget(widget)), equals(widget));
 
-    expect(exception, isInstanceOf<Error>());
+    expect(exception.message,
+        'Error happens when building Observer, but it was captured since disableErrorBoundaries==true');
```
pavanpodila, line 102: English should be written in past tense for this test

pavanpodila (changes requested):
Changelog entry and version upgrade must be made for this. Please run set-versions via npm as well

fzyzcjy:
Oh I forgot it... Will do it today

fzyzcjy:
Done. 

@pavanpodila May I know what is "npm set-version"? I have checked `docs/package.json`, google it, and search through repo without finding it.

fzyzcjy:
Done. @pavanpodila

pavanpodila (approved):

PR #784, opened 2022-04-02
Don't trigger unneeded ActionSpyEvent for AsyncAction

### LINKED ISSUE #734: Async action reports to spy hundreds of times on each run
Hi! We've added a spy for mobx so we can see in console what is happening with our stores. However it's not really usable for us because for some reasons async actions report to spy hundreds of times during a coarse of single action. I was able to reproduce this issue using [HackerNews](https://github.com/mobxjs/mobx.dart/tree/master/mobx_examples/lib/hackernews) example from mobx tutorial.

To reproduce:
1.  Replace HackerNewsStore in [news_store.dart](https://github.com/mobxjs/mobx.dart/blob/master/mobx_examples/lib/hackernews/news_store.dart) with the one that has async action instead of `FutureObserver`:
```
abstract class _HackerNewsStore with Store {
  final _hnApi = HNApi();

  @observable
  List<FeedItem> latestItems = [];

  @action
  Future<void> loadNews(FeedType type) async {
    if (type == FeedType.latest) {
      latestItems = await _hnApi.newest();
    } else if (type == FeedType.top) {
      topItems = await _hnApi.top();
    }
  }
```
2. Replace `FeedItemsView` from [news_widgets.dart](https://github.com/mobxjs/mobx.dart/blob/master/mobx_examples/lib/hackernews/news_widgets.dart) to use only `latestItems` from store for simplicity of example:
```
class FeedItemsView extends StatelessWidget {
  const FeedItemsView(this.store, this.type, {Key? key}) : super(key: key);

  final HackerNewsStore store;
  final FeedType type;

  @override
  // ignore: missing_return
  Widget build(BuildContext context) => Observer(builder: (_) {
        final List<FeedItem> items = store.latestItems;
        return ListView.builder(
            physics: const AlwaysScrollableScrollPhysics(),
            itemCount: items.length,
            itemBuilder: (_, index) {
              final item = items[index];
              return ListTile(
                leading: Text(
                  '${item.score}',
                  style: const TextStyle(fontSize: 20),
                ),
                title: Text(
                  item.title,
                  style: const TextStyle(fontWeight: FontWeight.bold),
                ),
                subtitle: Text('- ${item.author}'),
                onTap: () => store.openUrl(item.url),
              );
            });
      });
}
```
3. Enable Spying in [main.dart](https://github.com/mobxjs/mobx.dart/blob/master/mobx_examples/lib/main.dart):
```
mainContext.config = mainContext.config.clone(
    isSpyEnabled: true,
  );
```
4. Run codegen - `flutter packages pub run build_runner watch --delete-conflicting-outputs`
5. Run the project

**Expectation:**
`loadNews` action in HackerNewsStore runs only once and reports to spy about running once on Start and once on End.

**Reality:**
`loadNews` action in HackerNewsStore indeed runs only once (I can even test it by adding `print` statement inside async action) but reports to spy 450 (!) times about it's Start and End
image

Using:
flutter_mobx: 2.0.3
mobx: 2.0.6

s0nerik:
Fixes #734


fzyzcjy:
Could you please add a bit of tests to verify it

s0nerik:
> Could you please add a bit of tests to verify it :)

Yes! I've added a test that verifies the correct behavior. It breaks on master (eventCount == 4) and works with my patch (eventCount == 1).

I've also identified that [my previous guess](https://github.com/mobxjs/mobx.dart/issues/734#issuecomment-1086742966) about the reasons for extra events being reported might've been wrong. I couldn't reproduce the problem using "continuations" in the test environment, but easily did it with `scheduleMicrotask`.

fzyzcjy (approved):
LGTM

amondnet (approved):

pavanpodila (approved):

pavanpodila (changes requested):
Please add the Changelog, bump up the version with +2 on mobx. Make sure to invoke set-versions via npm or directly on terminal

s0nerik:
@pavanpodila please take a look. I've updated the version as you requested.

makstheimba:
Any progress on this? We are really waiting for this change as await is unusable without it

pavanpodila:
It's missing some changes to move forward

s0nerik:
@pavanpodila I've already updated the versions as you requested, twice. But since other commits get merged (often overriding the version you've requested me to adjust) and this one stays without attention, I'm not sure when I need to apply the changes to make the merge happen.

I simply don't have enough time to monitor every commit in this repository to apply  the version adjustment time and time again.

pavanpodila (approved):

pavanpodila:
@all-contributors add @s0nerik for code

amondnet:
@pavanpodila This pr failed tests.

https://github.com/mobxjs/mobx.dart/runs/6815659674?check_suite_focus=true


`dart run test test/async_action_test.dart -N "run allows updating observable values in an async function"`

```dart
00:00 +0 -1: AsyncAction run allows updating observable values in an async function [E]                                                                                                                     
  Expected: [0, 2, 3]
    Actual: [0, 3]
     Which: at location [1] is <3> instead of <2>
  
  package:test_api                  expect
  test/async_action_test.dart 42:7  main.<fn>.<fn>
  
00:00 +0 -1: Some tests failed. 
```

huybuidac:
@pavanpodila @s0nerik hey, can you revert this MR, @action with Future does not raise update UI after this MR

```dart
  @observable
  bool loading = false;

  @action
  login() async {
    loading = true; // **** does not update UI ****
    await Future.delayed(Duration(milliseconds: 1000));
    loading = false;
  }
```

Issue #803, opened 2022-05-06
Breaking change between mobx_codegen 2.0.4 and 2.0.6 ("late" and "context")


jfahrenkrug:
We noticed that when we had a dependency on `mobx_codegen` 2.0.4, it produced errors when running on Windows. The errors looked like this:

```type 'UnspecifiedInvalidResult' is not a subtype of type 'LibraryElementResult' in type cast```

There seems to be an issue in the `source_gen` package: https://github.com/dart-lang/source_gen/issues/557 These issues only happened on Windows, not on macOS.
However, when updating to `mobx_codegen` 2.0.6, the build_runner finishes successfully, BUT the generated code is invalid. It made every instance of `Atom` `late` and added a `context` parameter that would not compile. This would happen both on Windows and on macOS.

It would turn this:

```
final _$statusAtom = Atom(name: '_ConnectivityStoreBase.status');
```

into this:

```
late final _$statusAtom = Atom(name: '_ConnectivityStoreBase.status', context: context);
```

Our solution is to add `source_gen` as a direct dependency and to keep `mobx_codegen` at version `2.0.4`:

```
dev_dependencies:
  build_runner: ^2.1.10
  mobx_codegen: 2.0.4
  source_gen: 1.2.2
  flutter_test:
    sdk: flutter
```

However, we don't want to get stuck on version 2.0.4 forever :) What is the reason for the "stray" `late` and `context`s being added in 2.0.6? Is that a bug? Can it be turned off? Thanks!

**Comment by amondnet** [comment]
Use mobx ^2.0.7.

```yaml
dependencies:
  mobx: ^2.0.7
```

https://github.com/mobxjs/mobx.dart/pull/770
https://github.com/mobxjs/mobx.dart/blob/9c4eb1dbd99228cbc11425b59472f6142b3816fd/mobx/lib/src/api/store.dart#L11

jfahrenkrug:
@amondnet Thank you. I think I'm missing some context here (haha). What exactly changed and why is this necessary? Is there a place where I can read up on this? Thank you!!

amondnet:
@jfahrenkrug 

The MobX reactive system operates in a context where it is aware of the various observables and their linked reactions. Within this context, if an observable changes, any linked reaction would automatically execute. By default, there is a singleton, top-level context called the `mainContext`. All reactivity occurs within this context.

> Creating custom context is an advanced feature and not needed for your normal usage.

[...]

PR #806, opened 2022-05-11
Upgrading `mobx` to `2.0.7`

amondnet:
resolve https://github.com/mobxjs/mobx.dart/issues/803

Signed-off-by: Minsu Lee <amond@amond.net>

### DIFF
--- a/mobx_codegen/lib/mobx_codegen.dart
+++ b/mobx_codegen/lib/mobx_codegen.dart
@@ -2,4 +2,4 @@ library mobx_codegen;
 
 export 'src/mobx_codegen_base.dart';
 
-const version = '2.0.6';
+const version = '2.0.6+1';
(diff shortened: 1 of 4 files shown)


pavanpodila (approved):

pavanpodila (changes requested):
please run "set:versions" to set the mobx version as well. This is needed to set a const in mobx lib

mobx_codegen/lib/mobx_codegen.dart
```diff
@@ -2,4 +2,4 @@ library mobx_codegen;
 
 export 'src/mobx_codegen_base.dart';
 
-const version = '2.0.6';
+const version = '2.0.6+1';
```
amondnet, line 5: > please run "set:versions" to set the mobx version as well. This is needed to set a const in mobx lib

@pavanpodila done

PR #831, opened 2022-06-13
Rename asObservable() to obs() (primitive types only)

subzero911:
*`Fixes 2.0.7`

I want to push a little fix for my 2.0.7 update. It was very recent, and this fix won't touch many users.
I found in practice that `1.asObservable()` is ugly and has no big difference to `ObservableInt(1)`
So I changed it to `.obs()`
It is short and convenient for GetX users (which use `.obs` getter) to jump to MobX.
```dart
final x = 1.obs();
```


subzero911:
@pavanpodila can you approve the MR?

pavanpodila:
> @pavanpodila can you approve the MR?

there seems to be a big drop in coverage. Can you please look into that @subzero911  ?

subzero911:
`codecov/patch`
https://github.com/mobxjs/mobx.dart/pull/831/checks?check_run_id=6981802810
The check fails on lines 5, 12, 26, 33
These are function signatures - how am I supposed to test them? It seems to be impossible
Tests for function bodies are written.

Anyway, I added some tests for typedefs (to ensure they are corresponding to their generic types).

`codecov/project/flutter_mobx — 96.90% (-3.10%) compared to 4f98362`
I hadn't touch the `flutter_mobx` project - how could it drop the coverage?

mobx/lib/src/core/typedefs.dart
```diff
@@ -1,7 +1,7 @@
...
 
 /// aliases for primitive types
-typedef ObservableInt = Observable<int>;
-typedef ObservableDouble = Observable<double>;
-typedef ObservableBool = Observable<bool>;
-typedef ObservableString = Observable<String>;
+typedef ObsInt = Observable<int>;
+typedef ObsDouble = Observable<double>;
+typedef ObsBool = Observable<bool>;
+typedef ObsString = Observable<String>;
```
pavanpodila, line 7: I don't see a need for these typedefs...Its better to expose the raw generic type directly...more readable that way

pavanpodila (changes requested):

mobx/lib/src/core/typedefs.dart
subzero911: I agree, i will remove these typedefs.
They just make a code more tangled.

subzero911: @pavanpodila done.

gabrielscr: @pavanpodila please fix Observer, it's not reacting to any observables

subzero911: @gabrielscr how does it relate to my MR?
Create your own issue and ask for help there.

gabrielscr: @subzero911 no

pavanpodila (approved):

PR #839, opened 2022-07-05
Feat: Add simple version of type aliases and simple getter/setter

vursin:
Hello,

I think this package is so good but it has a bit of complexity when I always need to use `codegen` every time I have something change on the variables or methods. I found an effective way to use this plugin like:

``` dart
// Declare
final myVar = Observable('');

// My widget
Observer((_) => Text(myVar.value));

// Change the value
runInAction((_) => myVar.value = 'new value');

// Function
Future doSomething() async {
   runInAction((_) => myVar.value = 'wait');
   await Future.delayed(Duration(second: 3));
   runInAction((_) => myVar.value = 'done');
}
```

I also found it still has a bit of complication, so I create this PR to reduce the code like the below:
``` dart
// Declare
final myVar = ''.asObs; // Or ''.obs(), can also be 0.asObs, <String>[].asObs,...

// My widget
Observer((_) => Text(myVar()));

// Change the value
myVar('new value'); // This method also implement the runInAction in it.

// Function
Future doSomething() async {
   myVar('wait');
   await Future.delayed(Duration(second: 3));
   myVar('done');
}
```

I like the `.obs` but it's just used by the last PR.

---
## Pull Request Checklist

- [x] Update the **Changelog** to include all changes made in this PR
- [x] Run the **`set:versions` command** using `npm` or `yarn`. You can find this command in the `package.json` file in the root directory

### DIFF
--- a/mobx/CHANGELOG.md
+++ b/mobx/CHANGELOG.md
@@ -1,5 +1,33 @@
+## 2.0.7+5
+
+- Add simple version of type aliases for all possible types:
+
+``` dart
+final obsValue = 0.asObs; // = Observable<int>(0)
+final obsValue = ''.asObs; // = Observable<String>('')
+final obsValue = <String>[].asObs; // = ObservableList<String>();
+```
+
+- Add simple version of getter/setter for `Observable<T>` type:
+  - Get value: `final value = obsValue();`
+  - Set value: `obsValue('new value');`
+
+``` dart
+final obsValue = 'initial value'.asObs;
+
+// Get value instead of using `obsValue.value`
+print(obsValue()); // => print 'initial value'
+
+// Set value (Also wrapped with `runInAction` by default)
+// instead of using `runInAction(() => obsValue.value = 'new value');`
+obsValue('new value');
+print(obsValue()); // => print 'new value'
+```
+
 ## 2.0.7+4
+
 fixes:
+
 - shortened `1.asObservable()` to `1.obs()` (same for boolean, double, String) - [@subzero911](https://github.com/subzero911)
 - removed experimental typedefs from `2.0.7`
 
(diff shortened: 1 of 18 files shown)


amondnet:
@vursin 
https://mobx.netlify.app/api/context

> By default, it is set to observed. This means if there is an observable which is being observed and you mutate it outside of an action, MobX will throw an exception.

If you set `writePolicy` to `never`, you can mutate it outside of an action.

> /// Defines the behavior for observables mutated outside actions
> ///
> /// `observed`: If there are observers for the mutated observable, then throw. Else allow mutation outside an action.
> /// `always`: Always throw if an observable is mutated outside an action
> /// `never`: Allow mutating observables outside actions
> enum ReactiveWritePolicy { observed, always, never }

```dart
mainContext.config =
      ReactiveConfig(writePolicy: ReactiveWritePolicy.never);

// Declare
final myVar = Observable('');

// My widget
Observer((_) => Text(myVar.value));

// change the value
myVar.value = 'hello world';
```

Or you use custom context.
```dart
final myContext = ReactiveContext(
          config: ReactiveConfig(writePolicy: ReactiveWritePolicy.never));
      final myVar = ObservableString('a', context: myContext);

final dispose = autorun((_) {
        print(myVar.value);
      }, context: myContext);


myVar.value = 'new value1';
runInAction(() => myVar.value = 'new value2', context: myContext);
dispose();

// a
// new value1
// new value2
```

vursin:
@amondnet 
When I want to create a temporary reaction variable like `final isLoaded = false.asObs` in a StatelessWidget, I just need to call `isLoaded(true)` to change the `Observer` immediately, I think it's so convenient.

It also does not break the current rules for all other variables of the project, like I also want to use `isLoaded.value = true` in an `@action` method or throw an error when using `isLoaded.value = true` outside the `runInAction`.

Issue #840, opened 2022-07-05
ObservableFuture FutureState.pending state is not called


strmchsr:
ObservableFuture pending state is not called , FutureState.fulfilled is called directly.

I am using the following code:

```
@observable
ObservableFuture<HttpResponse<Reference?>>?  _responseFuture;

@computed
StoreState get stateResponse {
  if (_responseFuture == null ||
      _responseFuture?.status == FutureStatus.rejected) {
    return StoreState.initial;
  }
  // Pending Future means "loading"
  // Fulfilled Future means "loaded"

  return _responseFuture?.status == FutureStatus.pending
      ? StoreState.loading
      : StoreState.loaded;
}

@action
Future<bool?> getReferences(
    String mobile, {
      bool isResend = false,
    }) async {
  final future = _repository.someAPI("req");
  _responseFuture = ObservableFuture(future);
  // ObservableFuture extends Future - it can be awaited and exceptions will propagate as usual.
  try {
    var response = await _responseFuture;

    return true;
  }
  return false;
} catch (error) {

return false;
}
}
```
I am observing the stateResponse in UI to show the loader.

dylanalizon:
Same problem ... Any solution ?

strmchsr:
@dylanalizon had to remove @action to get it work

Issue #836, opened 2022-06-21
Observer not reacting to changes


ejdoming:
I'm invoking an action in form onchange event, and the action is doing ok. But the observer is not triggering the build method. Anyone have experienced this issue?

code in the widgets tree:
```
Form(
                key: _formKey,
                onChanged: () {
                  var ok = _formKey.currentState!.validate();
                  _vm.setFormOK(ok);
                },

....

Observer(
                        builder: (_) => MxButton(_vm.formOK.toString(),
                            _vm.formOK == true ? _vm.logar : null))

```
code in Store class:
```
  @observable
  bool formOK = false;
  @action
  void setFormOK(bool value) => formOK = value;

```
code in .g file (code behind)
```
@override
  bool get formOK {
    _$formOKAtom.reportRead();
    return super.formOK;
  }

  @override
  set formOK(bool value) {
    _$formOKAtom.reportWrite(value, super.formOK, () {
      super.formOK = value;
    });
  }

  late final _$_LoginPageVMActionController =
      ActionController(name: '_LoginPageVM', context: context);

  @override
  void setFormOK(bool value) {
    final _$actionInfo = _$_LoginPageVMActionController.startAction(
        name: '_LoginPageVM.setFormOK');
    try {
      return super.setFormOK(value);
    } finally {
      _$_LoginPageVMActionController.endAction(_$actionInfo);
    }
  }
```

[login_page.dart.txt](https://github.com/mobxjs/mobx.dart/files/8951655/login_page.dart.txt)
[login_page_vm.dart.txt](https://github.com/mobxjs/mobx.dart/files/8951656/login_page_vm.dart.txt)
[login_page_vm.g.dart.txt](https://github.com/mobxjs/mobx.dart/files/8951657/login_page_vm.g.dart.txt)

ejdoming:
New update for this issue. After upgrade, no Observer are triggering builder method anymore.

ejdoming:
I have solved the issue downgrading version of mobx to 2.0.7+2

quanghd96:
Same issue with version 2.0.7+3. I must downgrade to version 2.0.7

quanghd96:
It seems to only happen on some devices like virtual machines, some phones,...

amondnet:
#834

gabrielscr:
same problem here

gabrielscr:
I thoughted it was a problem with my app, but seens like's not.

gabrielscr:
[...]

Issue #834, opened 2022-06-16
async action behaves differently from 2.0.7+3.


amondnet:
1. example
    ```dart
    @observable
    String stuff = '';
    
    @observable
    loading = false;
    
    @action
    Future<void> loadStuff() async {
      loading = true;
      stuff = await fetchStuff();
      loading = false;
    }
    ```
2. test
    ```dart
    test('run allows updating observable values in an async function',
            () async {
          final action = AsyncAction('testAction');
    
          final counter = Observable<int>(0);
    
          final values = <int>[];
          autorun((_) {
            values.add(counter.value);
          });
    
          await action.run(() async {
            await sleep(10);
            counter
              ..value = 1
              ..value = 2;
            await sleep(10);
            counter.value = 3;
          });
    
          expect(counter.value, equals(3));
          expect(values, equals([0, 2, 3]));
       });
     ```

https://github.com/mobxjs/mobx.dart/pull/89

mobx: <= 2.0.7+2

1. example
    ```dart
    Future<void> loadStuff() async {
      runInAction(() {
        loading = true;
      })
      final _stuff = await fetchStuff();
      runInAction(() {
        stuff = _stuff;
        loading = false;
      });
    }
    ```
2. test
    values : [0, 2, 3]

https://github.com/mobxjs/mobx.dart/pull/784

mobx: >= 2.0.7+3

1. example
    ```dart
    Future<void> loadStuff() async {
      runInAction(() {
        loading = true;
        final _stuff = await fetchStuff();
        stuff = _stuff;
        loading = false;
      });
    }
    ```
2. test
    values : [0, 3]

https://github.com/mobxjs/mobx.dart/runs/6912223161?check_suite_focus=true

https://mobx.netlify.app/api/action/

> Changes to the observables are only notified at the end of the action. This ensures all mutations happen as an atomic unit. This also eliminates noisy notifications, especially if you are changing a lot of observables (say, in a loop).

https://mobx.js.org/actions.html#updating-state-using-actions

> They are run inside [transactions](https://mobx.js.org/api.html#transaction). No reactions will be run until the outer-most action has finished, guaranteeing that intermediate or incomplete values produced during an action are not visible to the rest of the application until the action has completed.

https://mobx.js.org/actions.html#asynchronous-actions

~~According to the docs, the `2.0.7+3` seems to be working correctly.~~

edited:

According to the docs, the `2.0.7+2` seems to be working correctly.

s0nerik:
[...]

PR #841, opened 2022-07-05
Fix: AsyncAction behavior

vursin:
Hello, 

I found this issue when I test my last PR. After doing some searching, I found that the PR #784 shouldn't be introduced without a breaking change version.

So I decided to fix this issue without breaking the code of the old users (who may be used this package since 2019 with PR #89) and allows anyone who wants to use new behavior (which is the official behavior of Mobx in the document).

I add an `asyncActionBehavior` flag to `ReactiveConfig` to control the behavior of `@action` on `Future` methods. This config is just needed for those who want to use new behavior to avoid breaking old users' code.

- New behavior:
``` dart
/// The same behavior with mobx >= 2.0.7+3
abstract class _YourStore with Store {
  @override
  ReactiveContext get context {
    return mainContext
      ..config.clone(asyncActionBehavior: AsyncActionBehavior.notifyOnlyLast);
  }
}
```

- Old behavior (currently users don't need to add this code): 
``` dart
/// The same behavior with mobx <= 2.0.7+2
abstract class _YourStore with Store {
  @override
  ReactiveContext get context {
    return mainContext
      ..config.clone(asyncActionBehavior: AsyncActionBehavior.notifyEachNested);
  }
}
```

Fixes: #834
Fixes: #836 
Fixes: #840

---
## Pull Request Checklist

- [x] Update the **Changelog** to include all changes made in this PR
- [x] Run the **`set:versions` command** using `npm` or `yarn`. You can find this command in the `package.json` file in the root directory

### DIFF
--- a/mobx/CHANGELOG.md
+++ b/mobx/CHANGELOG.md
@@ -1,3 +1,28 @@
+## 2.0.7+5
+
+- Add flag `asyncActionBehavior` to `ReactiveConfig` to control the behavior of `@action` on `Future` methods.
+
+``` dart
+/// Notify all nested actions in the `Future` method. This is the old behavior of the mobx <= 2.0.7+2
+/// This is currently the default behavior of the `@action` on `Future` methods.
+abstract class _YourStore with Store {
+  @override
+  ReactiveContext get context {
+    return mainContext
+      ..config.clone(asyncActionBehavior: AsyncActionBehavior.notifyEachNested);
+  }
+}
+
+/// Notify only the last action in the `Future` method. This is the new behavior of the mobx >= 2.0.7+3
+abstract class _YourStore with Store {
+  @override
+  ReactiveContext get context {
+    return mainContext
+      ..config.clone(asyncActionBehavior: AsyncActionBehavior.notifyOnlyLast);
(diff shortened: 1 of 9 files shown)


amondnet:
@vursin Thank you for your contribution. I think #784 should be reverted.

vursin:
@amondnet Additional information: This PR has the same behavior as reverted #784, just add an additional flag for those who want to continue using the new behavior.

ghenry:
Hi all,

Does this just need a final review before we can merge and close it?

Thanks,
Gavin.

amondnet:
@vursin Hello and thanks for the PR!

However, I'm not sure if we need this feature. Are there cases where this feature is needed? How often is that case? whole app, some `store` or some `method`?
Does this pr solve that case?

vursin:
@amondnet I think this feature is also not really needed myself but it will be better if there is a way to just notify when all the actions inside are finished.

I also think this is a better way to do that so feel free to close this PR.
> How about something like below?
> 
> ```dart
> @action(batch=true) //
> Future<void> runInBatch() {
> }
> ```

PR #861, opened 2022-09-05
`disableErrorBoundaries` will not only affect `Reaction`s, but also things like `Computed`s, so the original doc is wrong and should say `Derivation`s

fzyzcjy:
Describe the changes proposed in this Pull Request.

If the PR fixes a specific issue, reference the issue with **`Fixes #`**.

---
## Pull Request Checklist

- [ ] If the changes are being made to code, ensure the **version in `pubspec.yaml`** is updated. 
- [ ] Increment the **`major`/`minor`/`patch`/`patch-count`**, depending on the complexity of change
- [ ] Add the necessary **unit tests** to ensure the coverage does not drop
- [ ] Update the **Changelog** to include all changes made in this PR
- [ ] Run the **`set:versions` command** using `npm` or `yarn`. You can find this command in the `package.json` file in the root directory


pavanpodila (approved):

amondnet:
@fzyzcjy `melos run set_version` instead of `npm run set:versions`

pavanpodila (approved):

PR #876, opened 2022-10-15
Improved form validation example

asavchuk:
Improved form validation example now uses get_it as a dependecy injector, more real worl validation logic, serverside validation runs after local validation passed, inner page added.


pavanpodila (approved):

ghenry:
Hi all,

This looks good to merge and close too. Does the project want to recommend get_it or anything like that?

Thanks.

asavchuk:
> Does the project want to recommend get_it or anything like that?

not at all
there are many different possibilities and everyone can choose what is closer to him

pavanpodila (changes requested):
Can you update the docs as well with this example

asavchuk:
> Can you update the docs as well with this example

I'll try.

Issue #860, opened 2022-09-04
Add `@computedDisposable`, which automatically disposes the (heavy) computed data


fzyzcjy:
Example:

```
class A = _A with _$A;
class _A with Store {
  @computedDisposable
  MyHeavyObject get myObject => some_computation();
}
```

generates:

```
mixin _$AStore on _AStore, Store {
  Computed<bool>? _$myObjectComputed;

  @override
  bool get myObject =>
      (_$myObjectComputed ??= Computed(() => super.myObject,
              name: ...', disposeValue: (v) => v.dispose()) // NOTE 1
          .value;

  void dispose() {
    super.dispose();
    _$myObjectComputed.dispose(); // NOTE 2
  }
}
```

NOTE 1: Add this "disposeValue" automatically
NOTE 2: Add this "dispose" function

pavanpodila:
I prefer not to mix lifecycle management of resources with core observables. Computed never holds a value, but only depends on other value holding observables. So fundamentally it's bringing irrelevant responsibility. Would not vote for this.

You are bringing application level concepts here. Don't think it's the right place. Can you see if you can solve at the application layer?

fzyzcjy:
@pavanpodila Thanks for the reply! I agree and will not write code for the `@computedDisposable`. However, IMHO https://github.com/mobxjs/mobx.dart/issues/857 is still very much needed. (I guess your comment is only against `@computedDisposable`, not #857, but anyway I would write down some explanations)

The reason is that, sometimes the computed value is heavy. It can be a big controller that has a `dispose` method, a big object (memory-hungry) that needs to be disposed (freed) when no longer needed, etc. In such cases, it would be necessary to provide that `disposeValue` callback in #857. 

Indeed, I have tried to implement it without modifying mobx source code, but all failed. (If you are interested I can post my non-working solution here).

fzyzcjy:
@pavanpodila I love mobx (and use it extensively!), but I see Riverpod has some features that are so great - this is the main reason I am adding this (and planning to add others) - to make mobx have fewer shortcomings compared with riverpod! (Surely mobx has a lot of other strengths :) )

pavanpodila:
How about using an Atom to track the usage and then dispose when there are no more observers?

fzyzcjy:
LGTM, I will revert it in #844

PR #886, opened 2022-12-12
mobx utils

altynbek132:
added mobx utils
- obs from Listenable
- obs from ValueNotifier
- Stream of obs values

at least we should have ability to convert obs values into stream since streams are very useful for async world
https://github.com/mobxjs/mobx-utils#tostream
for example we could convert computed field to stream and solve #860 

added reporting manual change for observable and collections
there are some cases when you should need to report change, for example if using collections extensions (sortBy), or writing own function that converts some class similar to Listenable.

### DIFF
--- a/mobx/CHANGELOG.md
+++ b/mobx/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.1.4
+
+- add manual reporting change of observable and collections
+
 ## 2.1.2 - 2.1.3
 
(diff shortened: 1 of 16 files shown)


ghenry:
This looks very useful too. Anyone else have review and merge perms?

amondnet:
I also implemented some features of `mobx-utils`.
https://github.com/amondnet/mobx_utils

pavanpodila:
Coverage has dropped. Can we fix that @amondnet ?

Issue #551, opened 2020-08-14
Observer with excluded child branch


subzero911:
It seems like Observer rebuilds everything that is inside of him.
Consider adding `child: (_) => ...` parameter, for the child branch which is excluded from the rebuild. As in Consumer https://pub.dev/documentation/provider/latest/provider/Consumer-class.html

pavanpodila:
This is intended. You should put an `Observer` just at the places that use observables to minimize rebuilds. That also improves performance overall.

hacker1024:
I have a store that exposes a color. I use the color with a `ColorFiltered` widget, which applies a color filter using the color to its child.

I don't need the `ColorFiltered`'s child to be rebuilt every time the color changes, but there's no way to stop this happening at the moment.

subzero911:
@pavanpodila @hacker1024 hey I found some time to make an Observer with `child` parameter, take a look https://github.com/mobxjs/mobx.dart/pull/909

subzero911:
@hacker1024 just wrap your ColorFiltered's child into `Builder` widget, to stop it from rebuilding by Observer.

subzero911:
@pavanpodila that's something wierd:

In this test, child widget is not in Observer's immediate context and it is not being rebuilt
```dart
  testWidgets("Observer child not in immediate context did not change", (tester) async {
    final message = Observable('Click');
    final key = UniqueKey();

    await tester.pumpWidget(
      Observer(
        builder: (context) {
          return GestureDetector(
            onTap: () => message.value = 'Clicked',
            child: Builder(
              builder: (context) {
                return Text(message.value, key: key, textDirection: TextDirection.ltr);
              }
            ),
          );
        },        
      ),
    );

    expect(tester.widget<Text>(find.byKey(key)).data, equals('Click'));

    await tester.tap(find.byKey(key));
    expect(message.value, equals('Clicked'));

    await tester.pump();
    expect(tester.widget<Text>(find.byKey(key)).data, equals('Click')); // child text did not change
  });
```

But in this test, both Texts are rebuilt, despite one of them not in immediate context.

```dart
testWidgets("Observer child not in immediate context is not rebuilt", (tester) async {
    final message = Observable('Click');    
    final key1 = UniqueKey();
    final key2 = UniqueKey();

    await tester.pumpWidget(
      MaterialApp(
        home: Scaffold(
          body: Observer(
            builder: (context) {
              return Column(children: [
                ElevatedButton(
[...]

PR #909, opened 2023-03-24
Observer with excluded child

subzero911:
Sometimes it's nice to have an optional `child` parameter, so you can exclude child branch from re-render and improve performance.
This technique is used in AnimatedBuilder https://api.flutter.dev/flutter/widgets/AnimatedBuilder-class.html and Consumer https://pub.dev/documentation/provider/latest/provider/Consumer-class.html
I have added a `child` and `builderWithChild` parameters to Observer widget for such cases.
I've also added a test case for it.

Solves https://github.com/mobxjs/mobx.dart/issues/551

---
## Pull Request Checklist

- [x] Run the **`melo run set_version` command** from the root directory


subzero911:
I experimented more and I found that it's enough to put your subtree out of immediate context (using `Builder`, for example), and it will be excluded from rebuild. So we don't need this "child" parameter.
It's my bad, and I'll close this.
Gotta add a test showing this behaviour in a future PR.

amondnet:
Hi @subzero911, thank you for your contribution, but what are some examples of when this feature is needed?

subzero911:
@amondnet honestly, I can't come up with any real usecases for this feature 😄

subzero911:
It's crucial for AnimatedBuilder as it re-renders the widget every frame. And if I have some heavy child, I may want to exclude it from being re-rendered.
But Observer is mostly about one-shot rebuilds, it's not intended for animations or continious rebuilds.

PR #952, opened 2023-11-06
Observer.optimized

subzero911:
I already did a PR https://github.com/mobxjs/mobx.dart/pull/909 but I accidentally deleted the forked repo.
So I made it again and a little bit better.

Mostly you try to make Observer as small as possible. But sometimes you'll want to exclude the widget subtree from re-render and improve the performance.

Real world examples:
1) You have a colored Container with some heavy widgets inside (like `ListView` with `shrinkWrap: true`).
You want to change the color by Observable, but don't want to rebuild the whole ListView.
2) You have some fancy animation driven by Observable. You want to animate the widget and want to preserve its children from being rebuilt.

I introduce the new `Observer.optimized` which allows you to exclude the `child` branch.
Usage:
   ```dart
   Observer.optimized(
     builderOptimized: (context, child) => FooWidget(foo: foo, child: child),
     child: BarWidget(), // is not rebuilt
   ),
   ``` 

Solves https://github.com/mobxjs/mobx.dart/issues/551

@amondnet @fzyzcjy 

---
## Pull Request Checklist

### DIFF
--- a/flutter_mobx/lib/src/observer.dart
+++ b/flutter_mobx/lib/src/observer.dart
@@ -16,11 +16,7 @@ bool debugAddStackTraceInObserverName = true;
 /// See also:
 ///
 /// - [Builder], which is the same thing but for [StatelessWidget] instead.
-class Observer extends StatelessObserverWidget
-    // Implements Builder to import the documentation of `builder`
-    implements
-        // ignore: avoid_implementing_value_types
-        Builder {
+class Observer extends StatelessObserverWidget {
   // ignore: prefer_const_constructors_in_immutables
   Observer({
     Key? key,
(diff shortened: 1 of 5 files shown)


flutter_mobx/lib/src/observer.dart
```diff
@@ -16,39 +16,59 @@ bool debugAddStackTraceInObserverName = true;
...
           key: key,
           name: name,
           warnWhenNoObservables: warnWhenNoObservables,
         );
 
-  @override
-  final WidgetBuilder builder;
+  /// Observer which excludes the child branch
+  // ignore: prefer_const_constructors_in_immutables
+  Observer.optimized({
```
pavanpodila, line 38: We should call this `withChild` to make the intent clear. The idea is to only build parts that are actually changing and keeping rest of the tree intact. I think you would also need a `withChildren` variant as well which could be a `Map<String, Widget>`?. This variant will help in plugging in parts of the tree which are not changing. This is useful when you have more than one child.

pavanpodila (changes requested):

subzero911:
@pavanpodila done. I renamed `.optimized` to `.withChild` and `builderOptimized` to `builderWithChild`
I have no idea how to implement "withChildren" constructor to "exclude subtrees by map keys", and I don't think we actually need it. You can use a `Column(children: [...])` as a `child`, or you can use a composition of Observers.
I applied the same technique which was used in AnimatedBuilder, and `provider`'s `Consumer` and `Selector` - they all have the `child` parameter only. They don't have multiple children version. Check https://api.flutter.dev/flutter/widgets/AnimatedBuilder-class.html , "Performance optimizations" for details

pavanpodila (approved):

pavanpodila:
@all_contributors add @subzero911 for code

pavanpodila:
@all-contributors add @subzero911 for code

PR #987, opened 2024-03-21
build: melos

altynbek132:
Add mobx_lint_flutter_test package to melos.yaml
update melos dev dependency to version 5.2.1

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -17,7 +17,3 @@ scripts:
     exec: dart run $MELOS_ROOT_PATH/tool/expose_version.dart
   generate:
     run: cd mobx_examples && dart run build_runner watch --delete-conflicting-outputs
-
-command:
-  bootstrap:
-    usePubspecOverrides: true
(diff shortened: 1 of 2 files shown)


melos.yaml
```diff
@@ -16,7 +17,3 @@ scripts:
     exec: dart run $MELOS_ROOT_PATH/tool/expose_version.dart
   generate:
     run: cd mobx_examples && dart run build_runner watch --delete-conflicting-outputs
-
```
pavanpodila, line 19: This is required to work locally. We should not remove this

pavanpodila (changes requested):
Let's keep the command. Looks good otherwise

melos.yaml
altynbek132: this options has been removed in melos 3.1.1, isn't it?
https://github.com/invertase/melos/blob/melos-v3.1.1/docs/guides/migrations.mdx#L52

pavanpodila: Awesome .. didn't know that ...

pavanpodila (approved):

pavanpodila:
@all-contributors add @altynbek132 for code

PR #988, opened 2024-03-22
fix: stacktrace in computed

altynbek132:
when error is thrown in computed stacktrace should not be lost (check tests)
we have to use Error.throwWithStackTrace


mobx/test/exceptions_test.dart
```diff
@@ -26,4 +26,15 @@ void main() {
       expect(ex.stackTrace, isNotNull);
     }
   });
+
+  test('should preserve stacktrace', () async {
+    final stackTrace = StackTrace.current;
+    try {
+      Computed(() => Error.throwWithStackTrace(Exception('test'), stackTrace))
```
pavanpodila, line 33: Can we add a test with a simple exception instead of using `Error.throwWithStackTrace` ? Not all code will use this technique. We should see if the stacktrace is picked up automatically from deeper exceptions in code

mobx/lib/src/core/reaction.dart
```diff
@@ -91,7 +91,7 @@ class ReactionImpl with DebugCreationStack implements Reaction {
     }
```
pavanpodila, line 91: Also need to bump up version and add a changelog for this

pavanpodila (changes requested):

mobx/test/exceptions_test.dart
altynbek132: done

mobx/lib/src/core/reaction.dart
altynbek132: done

mobx/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: mobx
-version: 2.3.0+1
```
pavanpodila, line 2: The set_version command needs to be called from melos to also change the version.dart file for the package

pavanpodila (changes requested):

mobx/pubspec.yaml
pavanpodila: More details in this doc: https://mobx.netlify.app/development/release

altynbek132: done

altynbek132: formatting is in other pr, because there are several files being changed that are not related to current pr

PR #993, opened 2024-03-28
fix: Observable.value setter (covariant)

altynbek132:
Fix Observable.value setter when class with operator ==(covariant) is used as a value


pavanpodila (approved):
LGTM

PR #1029, opened 2024-12-05
withBuiltChild fix

subzero911:
I fixed `Observer.withBuiltChild` constructor.
I removed runtime asserts and unneeded nullables. So now it's neat.

I also added an example to the docs.

---
## Pull Request Checklist

### DIFF
--- a/docs/docs/api/observers.mdx
+++ b/docs/docs/api/observers.mdx
@@ -90,6 +90,35 @@ class _CounterExampleState extends State<CounterExample> {
 }
 ```
 
+#### Performance optimizations
+Mostly you try to make `Observer` as small as possible. But sometimes you want to exclude
+the widget subtree from being rebuilt and improve the performance.
+
+For that you can use an `Observer.withBuiltChild` constructor. It takes `builderWithChild`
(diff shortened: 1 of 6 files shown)


subzero911:
@pavanpodila @amondnet

subzero911:
image

@pavanpodila 
CI/CD says don't use explicit types for constructor super parameters. But I see nothing bad with it. You'd better set less strict linter rules.

Anyway, it doesn't relate to the current MR and should be fixed in a separate MR.

pavanpodila:
> image
> 
> 
> 
> @pavanpodila 
> 
> CI/CD says don't use explicit types for constructor super parameters. But I see nothing bad with it. You'd better set less strict linter rules.

I'm using default rules with no overrides. These are the ones recommended by Flutter itself

pavanpodila:
We should also upgrade the lints packages

subzero911:
image

It isn't shown as an error in the IDE, but CI/CD fails on this line for some reason.

subzero911:
Anyway, it's out of scope of the current MR. It should be done in a separate MR.

PR #1048, opened 2025-06-20
feat: addition of new method to collections to observe inner collection

CaLouro:
## Describe the changes proposed in this Pull Request.

This PR adds one new method to each observable collection that exposes the inner collection while triggering a observation report.

I've found the need of having such a method while working on a project and neither `nonObservableInner` or `toList()` would be appropriate. `nonObservableInner` would not report an observation and `toList()` would create a new list which for my need is unnecessary and could lead to performance problems on larger lists.

I propose the creation of a `inner` getter that exposes the inner collection while reporting the read.

## If the PR fixes a specific issue, reference the issue with **`Fixes #`**.

This PR does not closes any issues that I'm aware of.

---
## Pull Request Checklist

- [ ] Update the **Changelog** to include all changes made in this PR, organized by version
- [ ] Update the **docs** if there are any API changes or additions to functionality

### DIFF
--- a/mobx/lib/src/api/observable_collections/observable_list.dart
+++ b/mobx/lib/src/api/observable_collections/observable_list.dart
@@ -42,6 +42,13 @@ class ObservableList<T>
 
   List<T> get nonObservableInner => _list;
 
+  /// An observable version of [nonObservableInner].
+  List<T> get inner {
+    _context.enforceReadPolicy(_atom);
+    _atom.reportObserved();
+    return _list;
+  }
+
   Listeners<ListChange<T>>? _listenersField;
 
   Listeners<ListChange<T>> get _listeners =>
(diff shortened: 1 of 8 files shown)


CaLouro:
Hello everybody! @davidmartos96 @pavanpodila can someone take a look at this? Thank you all

CaLouro:
I see now that my implementation might not have been the best approach but seeing that the observable collections classes have already a `nonObservableInner` getter I thought that adding my `inner` implementation wouldn't break the architecture considerations.

Let me explain my scenario and maybe you could help me find another solution for my problem. Consider the code below:

```dart
final ObservableList<T> foo;

@computed
B get bar => Bar(foo);

@action
void baz() {
  // some list operation
}

```
Where `T` and `B` types are not important and I may have several `Actions` that manipulate the list.

This is the scenario that I have on my current code and at first I thought this setup would have worked. But it doesn't because the `foo` variable itself is not observed and only the inner list so the computed `bar` getter with a function call does not bind a dependency to `foo`. 

To bypass this limitation I'm currently doing the following 
```dart
@computed
B get bar => Bar(foo.toList());
```
and since my lists are expected to be small, I'm not finding performances issues but that could be a problem for bigger lists.

Do you know of a better way to approach this?

PR #1051, opened 2025-07-21
Update codegen to use new Element2 API from analyzer >= 7.4.0

### LINKED ISSUE #1039: [severe] incompatible to analyzer ^7.0.0
build runner even doesn't start
Image

### LINKED ISSUE #1049: incompatible with build_runner: ^2.6.0
Image

latest `build_runner` uses build 3.0.0 which is incompatible with current `mobx_codegen` version

josh-audio:
This PR updates `mobx_codegen` to support the latest `build` and `analyzer` packages, including:
- `analyzer: 7.7.0`
- `build: 3.0.0`
- `source_gen: 3.0.0`
- `build_test: 3.3.0`

For downstream consumers with custom codegen, the version bump will allow them to migrate their own code generators away from the recently deprecated element API in `package:analyzer`. I've also done the conversion for this package, so the cut-over to analyzer 8.0.0 (which fully removes the deprecated API) should be trivial.

Note that while analyzer 8.0.0 has been released, the latest `source_gen` does not yet support it, so this is the latest analyzer version we can upgrade to.

Fixes #1049, fixes #1039.

---
## Pull Request Checklist

### DIFF
--- a/mobx_codegen/lib/src/mobx_codegen_base.dart
+++ b/mobx_codegen/lib/src/mobx_codegen_base.dart
@@ -3,7 +3,7 @@
 
 import 'dart:async';
 
-import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/element2.dart';
 import 'package:analyzer/dart/element/nullability_suffix.dart';
 import 'package:analyzer/dart/element/type_system.dart';
 import 'package:build/build.dart';
@@ -43,7 +43,7 @@ class StoreGenerator extends Generator {
 
   Iterable<String> _generateCodeForMixinStore(
     LibraryReader library,
-    ClassElement baseClass,
+    ClassElement2 baseClass,
     TypeSystem typeSystem,
   ) sync* {
     final typeNameFinder = LibraryScopedNameFinder(library.element);
@@ -52,7 +52,7 @@ class StoreGenerator extends Generator {
       final mixedClass = otherClasses.firstWhere((c) {
         // If our base class has different type parameterization requirements than
         // the class we're evaluating provides, we know it's not a subclass.
-        if (baseClass.typeParameters.length !=
+        if (baseClass.typeParameters2.length !=
             c.supertype!.typeArguments.length) {
           return false;
         }
@@ -68,7 +68,7 @@ class StoreGenerator extends Generator {
       });
 
       yield _generateCodeFromTemplate(
-          mixedClass.name, baseClass, MixinStoreTemplate(), typeNameFinder);
+          mixedClass.name3!, baseClass, MixinStoreTemplate(), typeNameFinder);
       // ignore: avoid_catching_errors
     } on StateError {
       // ignore the case when no element is found
@@ -77,15 +77,15 @@ class StoreGenerator extends Generator {
 
   String _generateCodeFromTemplate(
     String publicTypeName,
-    ClassElement userStoreClass,
+    ClassElement2 userStoreClass,
     StoreTemplate template,
     LibraryScopedNameFinder typeNameFinder,
   ) {
     final visitor = StoreClassVisitor(
         publicTypeName, userStoreClass, template, typeNameFinder, options);
     userStoreClass
-      ..accept(visitor)
-      ..visitChildren(visitor);
+      ..accept2(visitor)
+      ..visitChildren2(visitor);
     return visitor.source;
   }
 }
--- a/mobx_codegen/lib/src/store_class_visitor.dart
+++ b/mobx_codegen/lib/src/store_class_visitor.dart
@@ -1,8 +1,8 @@
 // https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/doc/element_model_migration_guide.md
 // ignore_for_file: deprecated_member_use
 
-import 'package:analyzer/dart/element/element.dart';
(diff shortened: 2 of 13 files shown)


mobx_codegen/lib/src/mobx_codegen_base.dart
```diff
@@ -68,7 +68,7 @@ class StoreGenerator extends Generator {
       });
 
       yield _generateCodeFromTemplate(
-          mixedClass.name, baseClass, MixinStoreTemplate(), typeNameFinder);
+          mixedClass.name3!, baseClass, MixinStoreTemplate(), typeNameFinder);
```
josh-audio, line 71: The new `name3` from the Element2 API is nullable. This is inherited from the base `Element2`. In a number of places, I have made the assumption that the name in effect cannot be null, as in the case of this `ClassElement2`. This may be less safe than I think, or I may have made a mistake. I'm happy to change this approach.

mobx_codegen/lib/src/store_class_visitor.dart
```diff
@@ -182,24 +191,24 @@ class StoreClassVisitor extends SimpleElementVisitor {
         return;
       }
 
-      if (element.isAsynchronous) {
+      if (element.fragments.any((fragment) => fragment.isAsynchronous)) {
```
josh-audio, line 194: A `MethodElement2` can have multiple fragments (see [the migration guide](https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/doc/element_model_migration_guide.md) and [the augmentation libraries feature spec](https://github.com/dart-lang/language/blob/main/working/augmentation-libraries/feature-specification.md)). This can happen in the following scenario:

```dart
class MyClass {
  // This is the first fragment:
  Future<void> myMethod(int myArg);
}

// ...

augment class MyClass {
  // And this is the second fragment:
  augment myMethod(int myArg) async {
    // body here...
  }
}
```

Note that in this case only one of the fragments would actually be considered asynchronous - there is no `isAsyncrhonous` property on `MethodElement2`.

mobx_codegen/lib/src/template/method_override.dart
```diff
@@ -10,31 +10,32 @@ class MethodOverrideTemplate {
...
-    final param = (ParameterElement element) => ParamTemplate(
-        name: element.name,
+    final param = (FormalParameterElement element) => ParamTemplate(
+        name: element.name3!,
         type: typeNameFinder.findParameterTypeName(element),
         defaultValue: element.defaultValueCode,
         hasRequiredKeyword: element.isRequiredNamed);
 
-    final positionalParams = method.parameters
+    final positionalParams = method.formalParameters
```
josh-audio, line 23: Note the change to `formalParameters` in the migration doc:

https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/doc/element_model_migration_guide.md#formal-parameters

mobx_codegen/lib/src/type_names.dart
```diff
@@ -23,53 +23,63 @@ import 'package:mobx_codegen/src/template/comma_list.dart';
...
+        .expand((t) => t);
+
+    for (final import in libraryImports) {
+      final prefix = import.prefix2;
+      for (final entry in import.namespace.definedNames2.entries) {
+        _namesByElement[entry.value] = prefix?.name2 != null
+            // If the prefix exists, we prepend it to the name
+            ? '${prefix!.name2!}.${entry.key}'
+            // Otherwise, we just use the name
+            : entry.key;
```
josh-audio, line 55: A note on this change. The old code appeared to rely on `definedNames` including prefix. So in this case:

```dart
import 'dart:io' as io;
```

... all the names from `definedNames` would already include the prefix, as in `io.File`, `io.Directory`, and so on. This is no longer the case, so I am instead constructing the full element name manually.

mobx_codegen/test/test_utils.dart
```diff
@@ -41,28 +41,34 @@ Future<String> generate(String source) async {
...
+    return errors.join('\n');
+  }
+
+  final assetId = AssetId(pkgName,
+      '.dart_tool/build/generated/generator_sample/lib/generator_sample.g.dart');
+  final assetExists = await readerWriter.canRead(assetId);
+
+  if (!assetExists) {
+    return '';
+  }
```
josh-audio, line 60: This was not required with the old reader, but with the new reader, if a file is not present then it will throw if you try to access it. The tests appear to be expecting an empty string instead, so I am preserving that behavior here.

subzero911:
What I don't like about MobX is that maintainers tend to disappear. This PR has been hanging for over 3 weeks, and nobody still approved and merged it, while we all have to wait.

pavanpodila (approved):

josh-audio:
No problem, looks like a new deprecation from source_gen. Just pushed a fix.

ioma8:
Hello, can it be told when this will be released in pub package?

BirjuVachhani:
By the time this PR got merged and released, Analyzer v8 is out! :(

josh-audio:
@BirjuVachhani Analyzer V8 was out when I authored this PR as well. I did not try to incorporate it because it was not yet supported by `source_gen`. This did change before merge, which I did not notice. We can now support `analyzer` v8, so I've opened a PR:

https://github.com/mobxjs/mobx.dart/pull/1055

